PR #8, opened 2024-07-24
feat: Add decommutation of HK

mfacchinelli:
Main logic is in `imapProcessing.py`, where the processing step is chosen based on extension. In case of binary data, it is assumed it's HK.

The logic was pretty much copied from the SOC: [imap_processing/utils.py#L231](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/1e5692ba55cb66626605eee27e3db35ca30b5b7c/imap_processing/utils.py#L231)

Same goes for the utility function to convert to J2000 timestamp.


tests/test_main.py
```diff
@@ -35,11 +35,29 @@ def test_process_with_valid_config_does_not_error(tidyDataFolders):
...
         ],
     )
 
     print("\n" + str(result.stdout))
     assert result.exit_code == 0
     # check that file output/result.cdf exists
     assert Path("output/result.cdf").exists()
+
+
+def test_process_hk_ccsds(tidyDataFolders):
```
alastairtree, line 48: Can I tempt you to look at the [bit about tests naming in the dev standards doc](https://imperialcollege.atlassian.net/wiki/spaces/PMLSD/pages/429555802/Developer+Standards) - would be good to try and get everyone being consistent at all possible.

> Consider test name structure [function_or_feature_under_test]_does_[some_single_action]_when_[this_input_is_satisfied] 
> e.g. generate_reports_creates_plot_of_absolute_field_when_at_least_24h_of_science_data_is_available.

tests/test_main.py
```diff
@@ -35,11 +35,29 @@ def test_process_with_valid_config_does_not_error(tidyDataFolders):
...
+            "tests/config/hk.yaml",
+            "--file",
+            "MAG_HSK_PW.pkts",
+        ],
+    )
+
+    print("\n" + str(result.stdout))
+    assert result.exit_code == 0
+    # check that file output/result.cdf exists
+    assert Path("output/result.csv").exists()
```
alastairtree, line 63: Please can you add a test for headers and one that verifys an expected first row/last row/row count have actually been populated correctly

.devcontainer/devcontainer.json
```diff
@@ -61,6 +61,11 @@
                 "ms-python.python",
                 "ms-python.black-formatter",
                 "github.vscode-github-actions",
+                "actboy168.tasks",
+                "charliermarsh.ruff",
+                "ms-python.mypy-type-checker",
+                "eamodio.gitlens",
+                "GitHub.vscode-pull-request-github",
                 "ms-azuretools.vscode-docker"
```
alastairtree, line 69: Did you want veryone on the team to have all these plugins? Might be worth a chat - imaging at least some are valuable to all? I have found gitlens slow in the past so have previously avoided it but that may be old news.

alastairtree (changes requested):
This looks awesome! have added a couple of comments. Lets pair tomorrow on turning this into part of the "pipeline" demo

.devcontainer/devcontainer.json
mfacchinelli: I've only left the 2 that I think are useful for everyone. I've tried looking into how to have a personal set of extensions for devcontainers, but couldn't find anything useful. Will make do with re-installing the extensions whenever the container gets rebuilt.

tests/test_main.py
mfacchinelli: I've added a bit more context!

alastairtree (approved):

alastairtree:
Sorry I broke your PR by updating typer. Have fixed and regenerated the lock file. Will merge

PR #11, opened 2024-07-29
Add Calibration Toolkit

mhairifin:
Adds two commands to the IMAP MAG CLI that calibrate, and apply calibrations.

These use calibration tools from inside mag toolkit, packaged within the app to simplify integration for now.

Calibrations are not real, but basically form a skeleton inside of which real calibration and application could be performed.

### DIFF
--- a/src/mag_toolkit/calibration/CalibrationApplicator.py
+++ b/src/mag_toolkit/calibration/CalibrationApplicator.py
@@ -1,0 +1,46 @@
+import logging
+from pathlib import Path
+
+import numpy as np
+
+from src.mag_toolkit.calibration.CalibrationExceptions import CalibrationValidityError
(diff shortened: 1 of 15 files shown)


src/mag_toolkit/calibration/CalibrationApplicator.py
```diff
@@ -0,0 +1,46 @@
...
+
+        write_cdf(data, outputFile)
+
+        return outputFile
+
+    def checkValidity(self, data, calibrationCollection):
+        # check for time validity
+        if data.epoch[0] < np.datetime64(
+            calibrationCollection.valid_start
+        ) or data.epoch[1] > np.datetime64(calibrationCollection.valid_end):
```
mfacchinelli, line 44: Should `data.epoch[1]` be `data.epoch[-1]`? Maybe I don't know what datatype this `data` is in though!

PR #15, opened 2024-08-02
test: Add WireMock for mocking WebPODA and SCO web APIs

mfacchinelli:
Adding WireMock to WebPODA and SDC API tests. The main wiremocky file is `wiremockUtils.py`, which creates a `WireMockManager` class that can be used to set up WireMock for a specific test.

In the end, I've also had to refactor the `fetch-science` CLI method, as it was making a lot of unnecessary calls to the SOC.


.vscode/launch.json
```diff
@@ -1,5 +1,20 @@
 {
     "configurations": [
+        {
+            "console": "integratedTerminal",
+            "env": {
+                "TESTCONTAINERS_HOST_OVERRIDE": "host.docker.internal",
+                "WIREMOCK_DIND": "1"
+            },
```
alastairtree, line 8: tested it without these?

src/imap_mag/appConfig.py
```diff
@@ -18,17 +19,32 @@ class Source(BaseModel):
...
+
 class AppConfig(BaseModel):
     source: Source
     work_folder: Path = Path(".work")
     destination: Destination
     packet_definition: Optional[PacketDefinition] = None
+    api: Optional[API] = None
+
+    def __init__(self, **kwargs):
+        kwargs = dict((key.replace("_", "-"), value) for (key, value) in kwargs.items())
```
alastairtree, line 42:
```suggestion
		# replace hypens with underscores so that you can build config from constructor args, and still have them mapped to the hyphen split property names in the YAML files
        kwargs = dict((key.replace("_", "-"), value) for (key, value) in kwargs.items())
```

src/imap_mag/cli/fetchScience.py
```diff
@@ -0,0 +1,74 @@
...
+    __descriptors: list[str]
+    __flavors: list[str]
+
+    __data_access: ISDCDataAccess
+
+    def __init__(
+        self,
+        data_access: ISDCDataAccess,
+        descriptors: list[str] = ["norm", "burst"],
+        flavors: list[str] = ["-magi", "-mago"],
```
alastairtree, line 34: Switch these magic strings for strongly typed enums?

src/imap_mag/cli/fetchScience.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def download_latest_science(
+        self, **options: typing_extensions.Unpack[FetchScienceOptions]
+    ) -> list[Path]:
+        """Retrieve SDC data."""
+
+        downloaded = []
+
+        for descriptor in self.__descriptors:
+            date_range: pd.DatetimeIndex = pd.date_range(
```
alastairtree, line 50: This was from the other code but why do we use pandas here? Is it just for doing date ranges?

src/imap_mag/client/sdcDataAccess.py
```diff
@@ -76,16 +73,21 @@ def download(self, file_name: str) -> pathlib.Path:
...
     """Class for uploading and downloading MAG data via imap-data-access."""
 
-    def __init__(self, data_dir: str) -> None:
+    def __init__(self, data_dir: str, sdc_url: str | None = None) -> None:
+        """Initialize SDC API client."""
+
         imap_data_access.config["DATA_DIR"] = pathlib.Path(data_dir)
+        imap_data_access.config["DATA_ACCESS_URL"] = (
+            sdc_url or "https://api.dev.imap-mission.com"
+        )
```
alastairtree, line 85: Think I would be tempted to ditch the config object and the magic strings and just have properties on the `SDCDataAccess` object?

src/imap_mag/client/sdcDataAccess.py
```diff
@@ -97,45 +99,16 @@ def get_file_path(**options: Unpack[FileOptions]) -> tuple[str, str]:
...
 
     def upload(self, file_name: str) -> None:
-        """Upload data to imap-data-access."""
-
         logging.debug(f"Uploading {file_name} to imap-data-access.")
 
         try:
             imap_data_access.upload(file_name)
         except imap_data_access.io.IMAPDataAccessError as e:
             logging.warn(f"Upload failed: {e}")
```
alastairtree, line 107: This probably should not fail in this way as it is does not propagate the exception

src/imap_mag/main.py
```diff
@@ -181,7 +183,11 @@ def fetch_binary(
     packet: str = appUtils.getPacketFromApID(apid)
     logging.info(f"Downloading raw packet {packet} from {start_date} to {end_date}.")
 
-    poda = webPODA.WebPODA(auth_code, configFile.work_folder)
+    poda = WebPODA(
+        auth_code,
+        configFile.work_folder,
+        configFile.api.webpoda_url if configFile.api else None,
```
alastairtree, line 189: this should crash/error if the url is none and not pass in None

tests/test_main.py
```diff
@@ -96,10 +113,92 @@ def test_fetch_binary_downloads_hk_from_webpoda(tidyDataFolders):
...
+        json.dumps({}),
+    )
+    wiremock_manager.add_string_mapping(
+        "/query?instrument=mag&data_level=l1b&descriptor=burst-magi&start_date=20250502&version=latest&extension=cdf",
+        json.dumps({}),
+    )
+    wiremock_manager.add_string_mapping(
+        "/query?instrument=mag&data_level=l1b&descriptor=burst-mago&start_date=20250502&version=latest&extension=cdf",
+        json.dumps({}),
+    )
```
alastairtree, line 163: You can use wildcards in mapping URLs so all these mappings can become one base one that returns `{}` and one specific that returns `query_response`

alastairtree (changes requested):
Awesome. Thanks for this. Sorry I made a few review comments on the code that was my fault!

.vscode/launch.json
mfacchinelli: Yes, just did. They are not needed!

src/imap_mag/cli/fetchScience.py
mfacchinelli: Yes, that was the best function I found

src/imap_mag/cli/fetchScience.py
mfacchinelli: Good idea! Will do

src/imap_mag/client/sdcDataAccess.py
mfacchinelli: Not sure I get it... The config is how the SDC lets us configure the URL (we can also use env variables, but this is consistent with the WebPODA approach).

src/imap_mag/main.py
mfacchinelli: The url is `None` by default. If it's `None` we use the "real" address.

src/imap_mag/client/sdcDataAccess.py
alastairtree: Oh I see sorry, think this was me not understanding

src/imap_mag/cli/fetchScience.py
alastairtree: Feels like a big dependency to add for pretty simple utility. We have not really had the pandas vs numby vs both discussion but would be tempted to say we should not pull in all of pandas just for this. There might be some code you can inline from https://github.com/pandas-dev/pandas/blob/main/pandas/core/indexes/datetimes.py#L130 instead but I think I would just use list comprehension:

```
start = datetime.datetime(2024,8,1)
end = datetime.datetime(2024,8,7)
[end - datetime.timedelta(days=x) for x in range((end-start).days+1)]

# returns array. the +1 is inclusive of start and end:
[datetime.datetime(2024, 8, 7, 0, 0), datetime.datetime(2024, 8, 6, 0, 0), datetime.datetime(2024, 8, 5, 0, 0), datetime.datetime(2024, 8, 4, 0, 0), datetime.datetime(2024, 8, 3, 0, 0), datetime.datetime(2024, 8, 2, 0, 0), datetime.datetime(2024, 8, 1, 0, 0)]

```

mfacchinelli: I also use `pandas` for converting a string to a `datetime` ([`pandas.to_datetime`](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html)), as it does not require a specific input format for the date/time, but is flexible and can interpret all sorts of different formats. I think this function would be harder to replace than `pandas.date_range`, unless we enforce a (few?) specific format(s). 

I am not familiar with the differences between `pandas` and `numpy`, so I have no preference. I've had a quick try at removing `pandas` from our `pyproject.toml` dependencies, and we still end up pulling it in in our lock file, as some of our dependencies (including `numpy`??) depend on it...

alastairtree: Pandas depends on numpy, not the other way around, Maybe have a look at dateutil?

tests/test_main.py
mfacchinelli: Nice one!

alastairtree:
I have merged the other stuff to main, so it should be possible to rebase this on main and it can then be merged

PR #18, opened 2024-08-09
Task/tech debt

alastairtree:
This sorts out come of the stuff in https://github.com/ImperialCollegeLondon/mag-lab-data-platform/issues/29 and it closes https://github.com/ImperialCollegeLondon/mag-lab-data-platform/issues/31

@mfacchinelli please can you review?

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,5 +1,5 @@
 # DB Connection string
-SQLALCHEMY_URL=postgresql://USER_HERE:PASSWORD_HERE@db:5432/imap
+SQLALCHEMY_URL=postgresql+psycopg://USER_HERE:PASSWORD_HERE@db:5432/imap
 
 # API credentials fpr LASP PODA API - See https://lasp.colorado.edu/ops/imap/poda/
 WEBPODA_AUTH_CODE=
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -4,9 +4,13 @@ set -e
 IMAGE_NAME="${IMAGE_NAME:-imap-pipeline-core/imap-mag}"
 
 
+# docker run --rm -it \
+#   --entrypoint /bin/bash \
+#   --env-file dev.env \
+#   -v /mnt/imap-data:/data \
+#   $IMAGE_NAME
+
 docker run --rm -it \
-  --entrypoint /bin/sh \
   --env-file dev.env \
   -v /mnt/imap-data:/data \
   $IMAGE_NAME
-
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -1,16 +1,32 @@
 #!/bin/bash
 set -e
 
+# Example:
+# ./run-docker.sh
+# ./run-docker.sh -i   // for interactive debugging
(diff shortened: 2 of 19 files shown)


.env.template
```diff
@@ -1,5 +1,5 @@
 # DB Connection string
-SQLALCHEMY_URL=postgresql://USER_HERE:PASSWORD_HERE@db:5432/imap
+SQLALCHEMY_URL=postgresql+psycopg://USER_HERE:PASSWORD_HERE@db:5432/imap
```
mfacchinelli, line 2: Just for my own understanding, what does this `psycopg` do?

run-docker.sh
```diff
@@ -4,9 +4,13 @@ set -e
 IMAGE_NAME="${IMAGE_NAME:-imap-pipeline-core/imap-mag}"
 
 
+# docker run --rm -it \
```
mfacchinelli, line 7: What are the commented out lines for?

src/imap_db/migrations/env.py
```diff
@@ -1,10 +1,9 @@
 from logging.config import fileConfig
 
 from alembic import context
+from imap_db.model import Base
```
mfacchinelli, line 4: Maybe we should configure the import sorter to move our APIs in a separate block at the end. I preferred them being separate. But I can do that!

src/imap_mag/imapProcessing.py
```diff
@@ -34,10 +35,44 @@ class HKProcessor(FileProcessor):
     xtcePacketDefinition: Path
 
     def initialize(self, config: appConfig.AppConfig) -> None:
-        self.xtcePacketDefinition = config.packet_definition.hk
+        # first try the file path as is, then in the same directory as the module, then fallback to a default
+        pythonModuleRelativePath = Path(
```
mfacchinelli, line 39: This is something we might need to do in more places than just here. Could be a utility?

mfacchinelli (approved):
Looks good, thanks Alastair! Just some minor comments.

Great to see that we no longer need the `src` import everywhere 😄

.env.template
alastairtree: The scarily named Psycho PG is the postgres driver for python. This change is needed by the update from v2 to v3 - it seems to insist on the driver name being in the connection string now.

src/imap_db/migrations/env.py
alastairtree: That would be nice!

run-docker.sh
alastairtree: Oh that is just a tweak of the code that changes the entrypoint to bash which is very handy to debug the container because you then get an interactive shell inside the container rather than something running the entrypoint script. I will make something a bit nicer than a commented out block

src/imap_mag/imapProcessing.py
alastairtree: Agreed - when we have a use for it I will do the refactor!

PR #57, opened 2025-03-12
feat: Poll WebPODA and SDC

mfacchinelli:
## Open Questions

* What if a file exits in the database/file storage, but not in the other?
* HK data could be downloaded from WebPODA without splitting by day (could result in large BIN files).
* Do we really need to re-download from the progress timestamp, when it's really old?
* Should we store HK "L0" (i.e., raw packets) in a folder structure that follows ERT, instead of S/C time? Storing as S/C time does not make much sense, as one ERT day can contain any random S/C date (although unlikely in flight).

## Issues

* When downloading new HK data from a "progress timestamp", there will already be a file (say v000) containing the data downloaded on the previous day. The data downloaded on this new request will be saved in a new version of the same file (v001), but will only contain part of the data for that day, even though at this point, the whole day should be available. This can be fixed by:
    * Re-downloading the whole day, instead of starting from the "progress timestamp"
    * Stitching together the two files into v002 (can be done by a separate job/task that runs at a different time)
* S/C time in WebPODA is decoded slightly differently from how we do it, which results in a few second difference. WebPODA defines the time as UTC (see [doc](https://lasp.colorado.edu/ops/imap/poda/#:~:text=UTC%20time%20range%20in%20ISO%208601%20format)). 

### Examples

E.g., the files downloaded with the following 2 methods are different, even though they should be the same:

1) Download data received on 12 and 13 April 2025 (an MSIM3 day), which contains data from the S/C date 30/09/2026:
    ```
    poetry run imap-mag fetch binary --packet sid3_pw --start-date 2025-04-12 --end-date 2025-04-13 --ert
    poetry run imap-mag process output/imap/mag/hsk-pw-raw/2026/09/imap_mag_hsk-pw-raw_20260929_v000.pkts output/imap/mag/hsk-pw-raw/2026/09/imap_mag_hsk-pw-raw_20260930_v000.pkts
    ```
    (Note that we need to process both files corresponding to 29/09 and 30/09 as, since we download data with "ERT", it is not easy to give an accurate S/C date to the file, as each downloaded binary may contain multiple days!)
2) Download the S/C date 30/09/2026 directly:
    ```
    poetry run imap-mag fetch binary --packet sid3_pw --start-date 2026-09-30 --end-date 2026-09-30
    poetry run imap-mag process output/imap/mag/hsk-pw-raw/2026/09/imap_mag_hsk-pw-raw_20260930_v001.pkts
    ```

This will result in two CSV files:

1) First method: `output/2026/09/30/imap_mag_hsk-pw_20260930_v000.csv`
2) Second method: `output/2026/09/30/imap_mag_hsk-pw_20260930_v001.csv`

The first file contains two extra data points corresponding to the J2000 time 30/09/2026 00:00:02.


pyproject.toml
```diff
@@ -34,6 +34,8 @@ cdflib = "^1.3.1"
 psycopg = {extras = ["binary"], version = "^3.2.1"}
 prefect = {extras = ["shell"], version = "^3.1.7"}
 prefect-docker = "^0.6.1"
+psycopg2 = "^2.9.10"
+psycopg2-binary = "^2.9.10"
```
alastairtree, line 38: I dont think this should be necessary - we are supposed to be using v3 of the driver. https://www.psycopg.org/psycopg3/docs/basic/install.html
Suspect we went wrong with the connection string or something

src/imap_mag/DB.py
```diff
@@ -42,32 +64,96 @@ def __init__(self, db_url=None):
...
-            raise e
-        finally:
-            session.close()
+    @__session_manager
+    def get_download_progress_timestamp(self, item_name: str) -> datetime | None:
+        session = self.__get_active_session()
+        download_progress = (
+            session.query(DownloadProgress).filter_by(item_name=item_name).first()
+        )
+        self.update_download_progress(item_name, last_checked_date=datetime.now())
```
alastairtree, line 118: A "Get" operation doing an update is suprising - why?

src/imap_mag/DB.py
```diff
@@ -42,32 +64,96 @@ def __init__(self, db_url=None):
...
+        )
+
+        if download_progress is None:
+            download_progress = DownloadProgress(item_name=item_name)
+
+        if progress_timestamp is not None:
+            logger.info(
+                f"Updating progress timestamp for {item_name} to {progress_timestamp}."
+            )
+            download_progress.progress_timestamp = progress_timestamp
```
alastairtree, line 145: This logic should be pushed into the download progress entity so that the calling code looks something like this:
```
progress = db.get_download_progress(id)
progress.record_successful_download(timestamp)
db.save(progress)
```
Otherwise you end up in an anti-pattern known as [Anemic domain model](https://en.wikipedia.org/wiki/Anemic_domain_model)

src/imap_mag/DB.py
```diff
@@ -117,7 +201,7 @@ def add_file(
                 )
             )
         except Exception as e:
-            logging.error(f"Error inserting {destination_file} into database: {e}")
+            logger.error(f"Error inserting {destination_file} into database: {e}")
             destination_file.unlink()
```
alastairtree, line 205: Thanks for tidying up the logger!

src/imap_mag/api/apiUtils.py
```diff
@@ -93,17 +96,17 @@ def prepareWorkFile(file, configFile) -> Path | None:
     files.sort(key=lambda f: f.stat().st_mtime, reverse=True)
 
     if len(files) == 0:
-        logging.critical(f"No files matching {file} found in {folder}")
-        raise typer.Abort()
+        logger.critical(f"No files matching {file} found in {folder}")
+        raise FileNotFoundError(f"No files matching {file} found in {folder}")
```
alastairtree, line 100: Nice!

src/imap_mag/api/fetch/binary.py
```diff
@@ -22,25 +24,20 @@ def fetch_binary(
         ),
     ],
     apid: Annotated[int, typer.Option(help="ApID to download")],
-    start_date: Annotated[str, typer.Option(help="Start date for the download")],
-    end_date: Annotated[str, typer.Option(help="End date for the download")],
+    start_date: Annotated[datetime, typer.Option(help="Start date for the download")],
+    end_date: Annotated[datetime, typer.Option(help="End date for the download")],
```
alastairtree, line 28: Good!

src/prefect_server/pollHK.py
```diff
@@ -0,0 +1,148 @@
...
+
+    return f"Download-{apid_text}-from-{start_date.strftime('%d-%m-%Y')}-to-{end_date.strftime('%d-%m-%Y')}"
+
+
+@flow(
+    name=CONSTANTS.FLOW_NAMES.POLL_HK,
+    log_prints=True,
+    flow_run_name=generate_flow_run_name,
+)
+async def poll_hk_flow(
```
alastairtree, line 45: We need this logic/code to be testable and probably less dependent to prefect. Prefect server should be a "thin wrapper" around the core business logic and not a critical part of being able to exercise all parts of the system. Lets meet and work out how best to pull this out into it's own thing

alastairtree (commented):
Thanks - you have done loads of great stuff here! I reckon we should catch up next week as a couple of bit might be worth chatting through, but this is great progress.

src/imap_mag/DB.py
```diff
@@ -42,32 +64,96 @@ def __init__(self, db_url=None):
         # TODO: Check database is available
 
         self.engine = create_engine(db_url)
-        self.Session = sessionmaker(bind=self.engine)
+        self.session = sessionmaker(bind=self.engine)
+
+    @staticmethod
+    def __session_manager(func):
```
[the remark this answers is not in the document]
mfacchinelli: I made it up! I like the wrapper thing.

We can always not use the decorator if more specific handling is needed, but I thought that so far we have always done the same type of error handling, so it would be useful to reduce duplication.

mfacchinelli:
@alastairtree I think this is ready for a re-review!

src/prefect_server/prefectUtils.py
```diff
@@ -0,0 +1,127 @@
...
+    if not value:
+        logger.error(f"Block {secret_name} is empty.")
+        raise ValueError(f"Block {secret_name} is empty.")
+
+    logger.debug(f"Block {secret_name} retrieved successfully.")
+
+    return value
+
+
+def get_start_and_end_dates_for_download(
```
alastairtree, line 45: Firstly this function is not a prefectUtil - it is a an important part of the apps logic so this is in the wrong place i think.
Secondly, I have read it three times and I am really struggling to work out it actually does because it does too many things. Think we need to step back to what the users intention is, I think it is one of "Get the latest new data" or "Ensure we have all data between these 2 dates" or "Download all data between these 2 dates" and make it much more explicit what we are actually doing - like this is going to be really hard to maintain. And lastly a "get dates" method that saves to the database, and is actually starting the download process, is messy.  Think we need to step back to some OOP and follow the SRP. Model the system in classes, each class has a single responsibility, and each method does one thing only.

src/prefect_server/prefectUtils.py
```diff
@@ -0,0 +1,127 @@
...
+                    f"Packet {packet_name} is partially up to date. Downloading from {last_updated_date}."
+                )
+                packet_start_date = last_updated_date
+        else:
+            logger.info(
+                f"Not checking database and forcing download from {packet_start_date} to {packet_end_date}."
+            )
+
+    # Remove any timezone information
+    (packet_start_date, packet_end_date) = forceUTCTimeZone(
```
alastairtree, line 123: Maybe this should happen sooner and be a "convert" rather than "force"? When the application gets a datetime from a user, or from a datasource like a file, it should convert it into UTC immediately and then the application "thinks" in UTC only. There should never be a datetime in the DB that is not in UTC, otherwise we will get timezone issues.

src/imap_mag/appUtils.py
```diff
@@ -1,36 +1,41 @@
...
-APID_TO_PACKET = {
+APID_TO_PACKET: dict[int, str] = {
     1028: "MAG_HSK_SID1",
     1055: "MAG_HSK_SID2",
     1063: "MAG_HSK_PW",
     1064: "MAG_HSK_STATUS",
-    1082: "MAG_HSK_SID5",
-    1051: "MAG_HSK_SID11",
+    1082: "MAG_HSK_SCI",
+    # 1051: "MAG_HSK_SID11", SID11 is not supported by WebPODA
```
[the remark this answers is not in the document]
mfacchinelli: Should have used `MAG_HSK_PROCSTAT`!

mfacchinelli:
@mhairifin I think this is ready to be merged. There are some issues (see PR description), but they can probably be solved as separate tasks. This is humongous enough as it is!

mfacchinelli:
@mhairifin I think this is ready to be merged. There are some issues (see PR description), but they can probably be solved as separate tasks. This is humongous enough as it is!

Happy to go through this next week, as it is huge and might be good to give a demo anyway.

src/imap_mag/io/DatabaseFileOutputManager.py
```diff
@@ -0,0 +1,144 @@
...
+            return (metadata_provider.version, False)
+
+        preliminary_destination_file: Path = self.assemble_full_path(
+            Path(""), metadata_provider
+        )
+        database_file: File | None = self.__get_matching_database_file(
+            metadata_provider
+        )
+
+        while database_file is not None:
```
mhairifin, line 122: This seems like quite a convoluted way to find the most recent version - is there no way to query the database for all files with a certain prefix, and then sort the versions or something?

src/imap_mag/process/dispatch.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def dispatch(file: Path | list[Path], work_folder: Path) -> FileProcessor:
+    """Dispatch a file or a list of files to the appropriate processor."""
+
+    available_processor_types: list[type] = [
+        HKProcessor,
+    ]
+
+    if isinstance(file, list):
```
mhairifin, line 28: Should this instead be looping over these files to dispatch each of them individually, rather than just the first?

src/prefect_server/pollScience.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    if force_database_update and not force_ingestion_date:
+        logger.warning(
+            "Database cannot be updated without forcing ingestion date. Database will not be updated."
+        )
+
+    # If this is an automated flow run, use the database to figure out what to download,
+    # and use the ingestion date to download data.
+    automated_flow_run: bool = (start_date is None) and (end_date is None)
```
mhairifin, line 119: I think the automated_flow_run setting here has some unexpected effects, and it's not necessarily clear from the outside how this and force_database_update and force_ingestion_date interact

mhairifin (approved):
Looks good, great work! A few comments but I'm happy to let them be future tickets or work in progress to get this moved forward

mfacchinelli:
Tested once again on `mag-agent`. This is ready for merging!

PR #68, opened 2025-05-07
feat: Add support for SDC ingestion date and WebPODA ERT

mfacchinelli:


src/imap_mag/cli/fetchBinary.py
```diff
@@ -29,40 +38,76 @@ def __init__(
...
+    ) -> dict[Path, WebPODAMetadataProvider]:
         """Retrieve WebPODA data."""
 
-        downloaded: dict[Path, StandardSPDFMetadataProvider] = dict()
+        downloaded: dict[Path, WebPODAMetadataProvider] = dict()
+
+        if start_date == end_date:
+            start = start_date.replace(hour=0, minute=0, second=0, microsecond=0)
+            dates: list[datetime] = [start, start + timedelta(days=1)]
+        else:
```
mhairifin, line 54: I think this stretch of code here is a bit confusing - I can see what it's doing now but I think it could do with either a small refactor or at least a comment explaining what it's trying to achieve

src/imap_mag/api/fetch/science.py
```diff
@@ -33,46 +31,74 @@ def fetch_science(
     ],
     start_date: Annotated[datetime, typer.Option(help="Start date for the download")],
     end_date: Annotated[datetime, typer.Option(help="End date for the download")],
+    use_ingestion_date: Annotated[
```
mhairifin, line 34: From a UX perspective does it make more sense to have a flag here, or to e.g. set `--ingestion-start-date` and `--ingestion-end-date` instead? Don't have a strong opinion either way but maybe worth considering.

src/imap_mag/client/webPODA.py
```diff
@@ -49,41 +70,119 @@ def __init__(
...
+        packet: str,
+        start_date: datetime,
+        end_date: datetime,
+        ert: bool = False,
+    ) -> datetime | None:
+        logger.info(
+            f"Downloading time information for {packet} from {start_date} to {end_date}."
+        )
+
+        time_response: requests.Response = self.__download_from_webpoda(
```
mhairifin, line 156: This is more of a general question as I haven't been following this closely, but it seems to me that some of this code replicates or is similar to some of the functionality available in the `webpoda` module of `imap-data-access`. Does that not offer the right functionality for us, or is there another reason we are not using it?

src/imap_mag/io/OutputManager.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if not metadata_provider.supports_versioning():
+            logger.warning(
+                f"File {destination_file} already exists and is different. Overwriting."
+            )
+            return metadata_provider.version
+
+        while destination_file.exists():
+            logger.debug(
+                f"File {destination_file} already exists and is different. Increasing version to {metadata_provider.version + 1}."
```
mhairifin, line 79: Nit pick but since we are not actually checking the hash here, I am not sure it's necessarily true that the file is different

src/imap_mag/io/OutputManager.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+            metadata_provider.version += 1
+            updated_file = self.__assemble_full_path(metadata_provider)
+
+            if destination_file == updated_file:
```
mhairifin, line 84: Not sure I really understand the point of this conditional or how this scenario would come about

mhairifin (approved):
Couple questions but looks good to me overall!

src/imap_mag/api/fetch/science.py
mfacchinelli: It is a flag, it's defined as `--ingestion-date` a couple of lines later. The variable name is a little more explicit, but from command line it's off by default and is enabled with the flag.

src/imap_mag/io/OutputManager.py
mfacchinelli: Good point!

src/imap_mag/io/OutputManager.py
mfacchinelli: It's mostly to prevent infinite loops. If for some reason the variable does not change, we get stuck here forever (it happened once while I was testing it...). At least this breaks that from happening

mfacchinelli:
I think this is ready to be merged into my other branch (sorry @mhairifin, I'll ask you to review that one as well at some point) and for a demo on mag-agent!

PR #79, opened 2025-06-18
feat: Adding private calibration code

mhairifin:
Setting up the infrastructure to create private packages with the calibration code in them.

Can be deployed using: https://github.com/ImperialCollegeLondon/mag-lab-data-platform/pull/59


alastairtree (approved):
Extremely elegant solution. Nice work

PR #80, opened 2025-06-19
Feat/adding cal code

mhairifin:


alastairtree:
There is low code coverage on several of the new files so be good to keep working on test coverage in the next version and get it all back above 80

PR #87, opened 2025-06-27
fix: correct naming of offset files

mhairifin:
Add end dates to the names of offset files

Needed for SDC compatibility.

### DIFF
--- a/src/imap_mag/io/StandardSPDFMetadataProvider.py
+++ b/src/imap_mag/io/StandardSPDFMetadataProvider.py
@@ -21,6 +21,7 @@ class StandardSPDFMetadataProvider(IFileMetadataProvider):
     level: str | None = None
     descriptor: str | None = None
     content_date: datetime | None = None  # date data belongs to
+    end_date: datetime | None = None  # end date of validity for ancillary files
     extension: str | None = None
 
     def supports_versioning(self) -> bool:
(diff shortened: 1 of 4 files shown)


src/imap_mag/io/StandardSPDFMetadataProvider.py
```diff
@@ -21,6 +21,7 @@ class StandardSPDFMetadataProvider(IFileMetadataProvider):
     level: str | None = None
     descriptor: str | None = None
     content_date: datetime | None = None  # date data belongs to
+    end_date: datetime | None = None  # end date of validity for ancillary files
```
mfacchinelli, line 24: Technically, the end date is not part of the standard SPDF naming convention... https://imap-processing.readthedocs.io/en/latest/data-access/naming-conventions.html

Should we have a `AncillaryFileMetadataProvider`? It's a bit of a duplicate, but it does have a different format

mfacchinelli (changes requested):
Sorry, my comment kinda implies changing a lot of the code... I think the end date does not strictly belong in a standard SPDF file format.

src/imap_mag/io/AncillaryFileMetadataProvider.py
```diff
@@ -0,0 +1,97 @@
+import logging
```
mfacchinelli, line 1: Thanks Mhairi, that's great! Can you also add this file to the new `FileMetadataProviders` file? It might need a merge with `main` though

mfacchinelli (approved):
Thanks Mhairi!

src/imap_mag/io/AncillaryFileMetadataProvider.py
```diff
@@ -0,0 +1,132 @@
...
+            case "l2-calibration":
+                return Path("l2-rotation")
+            case "l1b-calibration":
+                return Path("l1b")
+            case _:
+                if self.descriptor.endswith("-offsets"):
+                    return Path("l2-offsets") / self.start_date.strftime("%Y/%m")
+                else:
+                    logger.error(
+                        f"Unknown descriptor '{self.descriptor}' for ancillary files. Defaulting to 'ancillary'."
```
mfacchinelli, line 53:
```suggestion
                        f"Unknown descriptor '{self.descriptor}' for ancillary files."
```

mfacchinelli (approved):
Just a small typo!

src/imap_mag/io/AncillaryFileMetadataProvider.py
mhairifin: Renamed all to PathHandlers - hope that's okay!

PR #95, opened 2025-07-03
task: HK and Science Metadata Providers

mfacchinelli:
This PR adds 2 new metadata providers:

* `HKMetadataProvider`: replaces `StandardSPDFMetadataProvider` for HK-specific use and incorporates `WebPODAMetadataProvider`, by including `ert`
* `ScienceMetadataProvider`: replaces `StandardSPDFMetadataProvider` for science-specific use and incorporates `SDCMetadataProvider`, by including `ingestion_date`

`StandardSPDFMetadataProvider` is still there, and it's the superclass for the 2 classes above. It provides definition of the properties and utilities for file naming.

`AncillaryFileMetadataProvider` no longer inherits from `StandardSPDFMetadataProvider`, as it does not need a `level` and `content_date` should really be `start_date`. There's probably an argument to say that `StandardSPDFMetadataProvider` is not the right name, as the folder structure of the HK and science providers no longer match that standard, but naming is horrible and I couldn't think of a better name.

I did not modify `calibrate` and `apply`, so these are to be updated, and so are their tests!

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -332,7 +332,7 @@ def test_get_filename_error_on_no_required_parameter(provider):
         ),
     ],
 )
-def test_suitable_provider_from_filename(filename, expected):
+def test_find_correct_provider_from_filename(filename, expected):
     actual = FileMetadataProviderSelector.find_by_path(
         filename, throw_on_none_found=False
     )
@@ -343,7 +343,7 @@ def test_suitable_provider_from_filename(filename, expected):
     "throw_error",
     [True, False],
 )
-def test_no_suitable_provider_from_filename(capture_cli_logs, throw_error):
+def test_behavior_on_no_suitable_provider_found(capture_cli_logs, throw_error):
     # Setup.
     path = Path("this-is_not_a_supported_file_v002.cdf")
 


tests/test_metadata.py
```diff
@@ -233,23 +160,192 @@ def test_find_provider_by_path(
...
+                extension="cdf",
+            ),
+        ),
+        (
+            "imap_mag_definitely_not_a_standard_spdf_file.txt",
+            None,
+        ),
+    ],
+)
+def test_suitable_provider_from_filename(filename, expected):
```
mhairifin, line 335: Can you rename this test - this doesn't just check suitable provider, it also checks the values. It is also testing two things - the find_by_path function and the decomposition of the file name into level descriptor etc. Should those be separate tests to not complicate debugging?

tests/test_metadata.py
```diff
@@ -233,23 +160,192 @@ def test_find_provider_by_path(
...
+    )
+    assert actual == expected
+
+
 @pytest.mark.parametrize(
     "throw_error",
     [True, False],
 )
-def test_no_suitable_provider(capture_cli_logs, throw_error):
+def test_no_suitable_provider_from_filename(capture_cli_logs, throw_error):
```
mhairifin, line 346: Not sure how useful of a test this is - only checks a file name that doesn't fit anything to make sure it doesn't fit anything, so not a particularly rigorous test. Think it might be more useful to make sure that an ancillary file name can't be picked up as a science file, and a calibration layer couldn't be mistaken for an ancillary file etc. That's partially what the test above does, but not really, as it relies on the ordering in the FileMetadataProviderSelector

mhairifin (approved):
Looks good! Couple notes on the tests

tests/test_metadata.py
mfacchinelli: This test just makes sure that the exception is thrown when it should be thrown and that the exception type is as expected. This is not for the functionality of the various providers, but for coverage really. Maybe the name is a bit misleading...

You are right thought that there aren't negative tests, but the test above makes sure that if the filename matches the right structure, it will be mapped onto the correct provider. So I think for now it is good enough

PR #106, opened 2025-07-09
refactor: HK download and file management

mfacchinelli:


src/imap_mag/process/HKProcessor.py
```diff
@@ -101,35 +128,70 @@ def process(self, files: Path | list[Path]) -> dict[Path, IFilePathHandler]:
...
+                existing_files
+            )
+            assert (apid in existing_data) and (len(existing_data) == 1)
+
+            return existing_data[apid]
+
+    def __load_and_decommutate_files(
+        self, files: list[Path]
+    ) -> dict[int, pd.DataFrame]:
+        combined: dict[int, pd.DataFrame] = dict()
```
alastairtree, line 179:
```suggestion
        l1_dataframe_by_apid: dict[int, pd.DataFrame] = dict()
```

src/imap_mag/process/HKProcessor.py
```diff
@@ -164,17 +226,27 @@ def __do_process(self, file: Path) -> dict[int, xr.DataArray]:
...
+    def __save_daily_data(
+        self, day: date, daily_data: pd.DataFrame, path_handler: HKPathHandler
+    ) -> tuple[Path, HKPathHandler]:
+        logger.debug(f"Generating file for {day.strftime('%Y-%m-%d')}.")
+
+        path_handler.content_date = datetime.combine(day, datetime.min.time())
+        file_path = self.__work_folder / path_handler.get_filename()
+
+        # Save to CSV.
+        daily_data.drop_duplicates(subset="shcoarse", keep="last", inplace=True)
```
alastairtree, line 247: You cannot dedupe by shcourse - it is perfectly valid to send 2 hk or event packets in the same second. You have to dedupe CCSDS by the triple (apid, shcourse, seq_count)

src/imap_mag/process/HKProcessor.py
```diff
@@ -164,17 +226,27 @@ def __do_process(self, file: Path) -> dict[int, xr.DataArray]:
...
+        daily_data.sort_index(inplace=True)
```
alastairtree, line 248: Sort by shcourse and then by seq counter

src/imap_mag/util/BinaryHelper.py
```diff
@@ -0,0 +1,83 @@
...
+from imap_mag.util.TimeConversion import TimeConversion
+
+
+class BinaryHelper:
+    """
+    Class to handle binary data processing.
+    """
+
+    @staticmethod
+    def get_apids_and_days(binary_file: Path) -> dict[int, set[date]]:
```
alastairtree, line 19: How about an OO version instead: `CCSDSBinaryPacketFile(file)` so you can reuse the packet definition?
and then method `get_apids_and_days()`

alastairtree (commented):
Sorry I think it would be worth a bit more time on this?

src/imap_mag/process/HKProcessor.py
```diff
@@ -71,28 +69,57 @@ def process(self, files: Path | list[Path]) -> dict[Path, IFilePathHandler]:
...
+        # Load binary files and extract dates.
+        apids_by_day: dict[int, set[date]] = dict()
 
-        for file in track(files, description="Processing HK files..."):
-            file_results: dict[int, xr.DataArray] = self.__do_process(file)
+        for file in files:
+            file_apids: dict[int, set[date]] = BinaryHelper.get_apids_and_days(file)
+
+            for apid, days in file_apids.items():
+                apids_by_day.setdefault(apid, set()).update(days)
```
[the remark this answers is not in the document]
mfacchinelli: Good idea!

src/imap_mag/process/HKProcessor.py
mfacchinelli: Good point!

src/imap_mag/io/InputManager.py
```diff
@@ -15,31 +16,51 @@ class InputManager:
     def __init__(self, location: Path) -> None:
         self.location = location
 
+    def get_all_file_versions(
```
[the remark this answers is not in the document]
mfacchinelli: Yes, I was being a bit lazy here! How about using `sequence` instead of `version`, then path handlers can use it as they like. Both part numbers and versions are essentially counters that increase, i.e., sequences, with different connotations based on context (which is defined by the path handler).

src/imap_mag/cli/cliUtils.py
```diff
@@ -45,6 +46,20 @@ def throw_error_file_not_found(source_folder: Path, filename: str) -> None:
...
+    file: Path, work_folder: Path, *, throw_if_not_found: Literal[True]
+) -> Path:
+    pass
+
+
+@overload
+def fetch_file_for_work(
+    file: Path, work_folder: Path, *, throw_if_not_found: Literal[False]
+) -> Path | None:
+    pass
```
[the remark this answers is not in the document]
alastairtree: > They do not look like a cliUtil?
Still think these are in the wrong place?

src/imap_mag/io/CalibrationLayerPathHandler.py
```diff
@@ -91,6 +86,6 @@ def from_filename(
                 instrument="mag",
                 calibration_descriptor=match["descr"],
                 content_date=datetime.strptime(match["date"], "%Y%m%d"),
-                version=int(match["version"]),
+                sequence=int(match["sequence"]),
                 extension=match["ext"],
```
alastairtree, line 90: I know this is annoying but i think files with `thing_v000.ext` have a version. but we have other files which are "parts" of a file - they should not have v in the file name - maybe they just have `thing_000.ext` or even `thing_part000.ext`. And I think the code should have both version and sequence/part listed, not just replacing all versions with sequence - version should still exist? Or have I misunderstood what you have done?

alastairtree: Or what about a general purpose solution - any file can have versions and parts? so `thing_v000.ext` thing_v000_part001.ext and both valid?

src/imap_mag/process/HKProcessor.py
```diff
@@ -164,17 +280,40 @@ def __do_process(self, file: Path) -> dict[int, xr.DataArray]:
...
+
+        path_handler.content_date = datetime.combine(day, datetime.min.time())
+        file_path = self.__work_folder / path_handler.get_filename()
+
+        # Save to CSV.
+        daily_data.drop_duplicates(
+            subset=[
+                CONSTANTS.CCSDS_FIELD.APID,
+                CONSTANTS.CCSDS_FIELD.SHCOARSE,
+                CONSTANTS.CCSDS_FIELD.SEQ_COUNTER,
```
alastairtree, line 307: Nice!

src/imap_mag/util/TimeConversion.py
```diff
@@ -46,3 +46,12 @@ def convert_j2000ns_to_date(
...
         ]
+
+    @staticmethod
+    def convert_met_to_date(
+        met: npt.ArrayLike,
+        reference_epoch: np.datetime64 = CONSTANTS.IMAP_EPOCH,
+    ) -> list[date]:
+        """Convert mission elapsed time (MET) to Python date."""
+        j2000ns = TimeConversion.convert_met_to_j2000ns(met, reference_epoch)
+        return TimeConversion.convert_j2000ns_to_date(j2000ns)
```
alastairtree, line 57: Should we add a comment to this that it does not use SPICE so is not accurate enough for science? And I assume this does not handle the leap seconds thing (which means we get different results to the SDC??)

src/imap_mag/io/CalibrationLayerPathHandler.py
mfacchinelli: What I had in mind was having a generic "sequence" that can be used as both part or sequence depending on context. But I realized that it does not translate well when searching for files, as one would be allowed to get the latest "part", which does not mean much...

So I've re-written the HK path handler to support parts, when L0, and versions when L1. This is yet another big change!

PR #114, opened 2025-07-28
Fix/timestamp precision bug

mhairifin:
Change internal representation of datetimes for calibration science to np.datetime64 to prevent bug with nanosecond precision.

Also changes read and write of cdfs to use xarray as the primary mode of reading and writing. While a slightly more complex interface, it preserves the full nanosecond specificity of the timestamps.

Additional changes mostly to preserve some ISTP compliance enforced by xarray CDF write.

Also added tests to ensure that we both read and write CDFs and JSON layers to full nanosecond specificity, and calibrate to nanosecond specificity.


src/mag_toolkit/calibration/CalibrationDefinitions.py
```diff
@@ -28,21 +39,41 @@ class Mission(str, Enum):
     IMAP = "IMAP"
 
 
-class CalibrationMetadata(BaseModel):
+def convert_time(value: str) -> np.datetime64:
+    return np.datetime64(value, "ns")
+
+
+def convert_timedelta(value) -> np.timedelta64:
```
alastairtree, line 46: Can't see this new function convert_timedelta used anywhere?

src/mag_toolkit/calibration/CalibrationLayer.py
```diff
@@ -24,25 +27,59 @@ def _write_to_csv(self, filepath: Path, createDirectory=False):
...
+                "valid_start_datetime": self.validity.start,
+                "valid_end_datetime": self.validity.end,
+            },
+            coords={"axis": ["x", "y", "z"]},
+            attrs=skeleton_cdf.attrs,
+        )
+
+        offsets_dataset.attrs["Generation_date"] = str(np.datetime64("now"))
+        offsets_dataset.attrs["Data_version"] = self.version
+
```
alastairtree, line 82: A bunch of magic strings in here that should be a constants so we dont have to change them in many different places and enforces compile time tooling on their references

src/mag_toolkit/calibration/ScienceLayer.py
```diff
@@ -117,34 +154,41 @@ def _from_csv(cls, path: Path):
...
+        raw_epoch = dataset["epoch"].values
+        epoch = cdfepoch.to_datetime(raw_epoch)
 
         if data is None or epoch is None:
             raise ValueError("CDF does not contain valid data")
 
         validity = Validity(start=epoch[0], end=epoch[-1])
 
-        sensor = Sensor.MAGO if cdf_loaded.attrs["is_mago"] else Sensor.MAGI
+        sensor = Sensor.MAGO if dataset.attrs["is_mago"][0] == "True" else Sensor.MAGI
```
alastairtree, line 170: Pydantic has a nice bool parser you could use, see https://stackoverflow.com/a/71510434/3140853

src/mag_toolkit/calibration/ScienceLayer.py
```diff
@@ -117,34 +154,41 @@ def _from_csv(cls, path: Path):
...
+                range=datapoint[3],
+            )
+            for raw_epoch_val, epoch_val, datapoint in zip(raw_epoch, epoch, data)
         ]
 
         return ScienceLayer(
-            id=cdf_loaded.attrs["Logical_file_id"][0],
-            mission=cdf_loaded.attrs["Mission_group"][0],
+            id=dataset.attrs["Logical_file_id"][0],
+            mission=dataset.attrs["Mission_group"][0],
```
alastairtree, line 191: Any way to reduce the duplication between `_from_cdf` and `_from_csc` - lots of same boilerplate there? Also one creates the class using `ScienceLayer` and the other use `cls` which is just a bit inconsistent?

src/mag_toolkit/calibration/ScienceLayer.py
mfacchinelli: Cool!

resource/l2_dsrf_skeleton_old.cdf
alastairtree, line 1: Delete this?

PR #115, opened 2025-07-30
task: Add creation, modified and deletion dates for database files

mfacchinelli:


src/imap_db/migrations/versions/2025_07_30-abd9fd48d848_added_creation_modified_and_deleted_.py
```diff
@@ -0,0 +1,36 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column("files", "date", nullable=False, new_column_name="content_date")
+    op.add_column("files", sa.Column("creation_date", sa.DateTime(), nullable=False))
+    op.add_column(
+        "files", sa.Column("last_modified_date", sa.DateTime(), nullable=False)
+    )
```
alastairtree, line 25: Good practise if the database also enforces the timestamps automatically, see https://stackoverflow.com/a/33532154/3140853 for ways to do this

Maybe this would work?
```
from sqlalchemy.sql import func

    op.add_column("files", sa.Column("creation_date", sa.DateTime(), nullable=False, server_default=func.now()))
    op.add_column(
        "files", sa.Column("last_modified_date", sa.DateTime(), nullable=False, onupdate=func.now())
    )
```

src/imap_mag/io/DatabaseFileOutputManager.py
```diff
@@ -72,7 +73,13 @@ def add_file(self, original_file: Path, path_handler: T) -> tuple[Path, T]:
...
                         size=destination_file.stat().st_size,
-                        date=path_handler.content_date,
+                        content_date=path_handler.content_date,
+                        creation_date=datetime.fromtimestamp(
+                            destination_file.stat().st_ctime
+                        ),
+                        last_modified_date=datetime.fromtimestamp(
+                            destination_file.stat().st_mtime
+                        ),
                         software_version=__version__,
```
alastairtree, line 83: This logic should be encapsulated in the `File` object constructor - File(destination_file, version, hash, content_date) and the rest should be internal to the class?

src/imap_mag/io/DatabaseFileOutputManager.py
```diff
@@ -72,7 +73,13 @@ def add_file(self, original_file: Path, path_handler: T) -> tuple[Path, T]:
                         version=path_handler.get_sequence(),
```
alastairtree, line 73: `version=path_handler.get_sequence()` - is this right?

alastairtree (commented):
Couple of small things but look good thanks!

src/imap_mag/io/DatabaseFileOutputManager.py
mfacchinelli: Good idea, it could be a helper method!

src/imap_mag/io/DatabaseFileOutputManager.py
mfacchinelli: Yes, that's the method for path helpers that support some kind of sequencing. Can be version or part number.

PR #127, opened 2025-08-05
Move off of wiremock fork

mhairifin:
Quick fix to move off of our patched wiremock fork, as the patch is now in the latest wiremock release.


alastairtree (approved):
Awesome!

PR #129, opened 2025-08-08
fix: MATLAB tests and SQL injection

mfacchinelli:


tests/test_calibration.py
```diff
@@ -397,7 +400,7 @@ def test_gradiometry_calibration_layer_is_created_with_correct_offsets_for_one_v
     assert grad_layer["method"] == "gradiometer"
     assert len(grad_layer["values"]) == 99
     assert datetime.strptime(
-        grad_layer["values"][0]["time"], format
+        grad_layer["values"][0]["time"][:-3], format
```
mhairifin, line 403: Instead of slicing off the last three digits, these datetimes should be parsed with np.datetime64 instead and compared as np datetime64 values I think

PR #130, opened 2025-08-08
task: (temp) remove win tests & pyinstaller, build only on PRs

alastairtree:


alastairtree:
Build for this is failing on the tests  - but those are also failing on main now, so unrelated (though they passed when merged so could be a timestamp issue mhairi mentioned. Either way will merge now, and then let #129 fix the tests

PR #131, opened 2025-08-12
fix: add additional test assertions for correct metadata

mhairifin:
Tests were not testing for the correct science in the layer files. Added tests to ensure this does not fail again.

Needs to also include an update to the calibration code to match an updated version.

https://github.com/ImperialCollegeLondon/IMAP_MAG_Experimental_Calibration/pull/5


mhairifin:
Sorry, just updated the tag!

mfacchinelli (approved):

PR #132, opened 2025-08-13
task: Split layer data into CSV

mfacchinelli:
This change introduces a new file for calibration, which contains layer data. I.e., the calibration values are no longer defined in the JSON layer, but in a separate CSV file.

> [!Note]
> Suggestions on names for new properties, classes, descriptors are welcome. I've tried to be consistent with Mhairi's nomenclature.

## Model Changes
This new CSV file is tracked in a metadata property for the `LayerMetadata` class, called `data_filename`.

When loading a `Layer` file (could be a science layer or calibration layer), if the `data_filename` property is empty, it is ignored and the data is loaded from the JSON file only. If the property is set, the data is loaded from the CSV file (and a warning is thrown if the data is defined in both - CSV files are prioritized). 

`CalibrationLayer` now supports loading from CSV, which of course is needed to enable this whole process.

## API Changes
The API signatures are **unchanged**. When passing in a calibration layer during `apply`, we read the JSON file and the CSV file is loaded and copied to the `work` folder if it's present, otherwise the property is ignored. This is the part of the code that is a bit "ugly" and complicated, but it does the job well: 
https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/f26990a0caccd0fa55620cac5c773964ee014031/src/imap_mag/cli/apply.py#L34-L113

For `calibrate`, there is a new argument that needs to be passed to the `run_calibration` methods, which is this new `datafile`, which is then passed on to MATLAB:
https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/f26990a0caccd0fa55620cac5c773964ee014031/src/mag_toolkit/calibration/calibrators/CalibrationJob.py#L101-L103

## File Names
Now that the data can be split into 2 files, we need 2 new file name patterns to handle them differently:

* Metadata file (i.e., the JSON file we already had), adds a `-meta` specifier: e.g., `imap_mag_noop-layer-meta_20250814_v001.json`
* Data file (i.e., the new CSV file), adds a `-data` specifier: e.g., `imap_mag_noop-layer-data_20250814_v001.csv`

There's 2 things I don't really like:

1. The "metadata" file contains a property that is called `metadata`, making it a bit redundant and confusing. What could be a better name though? It does contain a lot of metadata, but still has the possibility of holding "values".
2. "data" and "metadata" are quite similar names, so I've tried to reduce confusion by naming one `...-meta` and the other `...-data`.

To handle the new names, there's 2 new path handlers, which inherit from the existing `CalibrationLayerPathHandler`, and customize the descriptor and extension of the calibration layer.

## MATLAB Changes
See this PR: https://github.com/ImperialCollegeLondon/IMAP_MAG_Experimental_Calibration/pull/6

The changes there match what's been changed here.

## Testing
I've reorganized a bit the files in the `datastore` folder, only keeping files that match the datastore requirements (in terms of naming) and removed edge cases files (they are now defined in the `test_data` folder, and get copied over when needed in a test).

I've separated the `apply` tests into its own file (`test_apply.py`), as I've added a few more and `test_calibration,py` was getting big. I've also added a few more tests in `test_layers.py` to cover the changes in `CalibrationLayer`.

In the MATLAB PR, you will see there's now also MATLAB tests! ❤️


src/mag_toolkit/calibration/calibrators/EmptyCalibration.py
```diff
@@ -43,8 +44,10 @@ def _get_path_handlers(self, calibration_job_parameters: CalibrationJobParameter
 
         return path_handlers
 
-    def run_calibration(self, calfile, config=None):
-        # produce an epmpty calibration through matlab
+    def run_calibration(
+        self, calfile: Path, datafile: Path, config=None
```
alastairtree, line 48: These 2 paths are so tightly coupled I think they should be one thing

src/mag_toolkit/calibration/CalibrationLayer.py
```diff
@@ -90,3 +97,82 @@ def _write_to_cdf(self, filepath: Path, createDirectory=False) -> Path:
...
+                and calibration_metadata_handler.calibration_descriptor
+            )
+            else CalibrationMethod.NOOP
+        )
+
+        return cls(
+            id="",
+            mission=Mission.IMAP,
+            validity=validity,
+            sensor=Sensor.MAGO,
```
alastairtree, line 167: Do we need/mean to hardcode to mago here?

src/mag_toolkit/calibration/Layer.py
```diff
@@ -22,11 +25,32 @@ class Layer(BaseModel, ABC):
...
+            return model
+
+        files_to_try: list[Path] = [
+            model.metadata.data_filename,
+            path.parent / model.metadata.data_filename.name,
+        ]
+
+        for file in files_to_try:
+            if file.exists():
+                return cls._load_data_file(file, model)
```
alastairtree, line 45: Why is this on the class? Couln't it just be an instance method called once you have instantiated a layer object? Would be simpler and no need to pass around model object?

tests/datastore/calibration/layers/2025/10/imap_mag_noop-layer-data_20251017_v001.csv
```diff
@@ -0,0 +1,17 @@
+time,offset_x,offset_y,offset_z,timedelta,quality_flag,quality_bitmask
+2025-10-17T02:11:51.521309000,0,0,0,0,0,0
```
alastairtree, line 2: More general qs for @mhairifin - should layers have time, epoch or both?

alastairtree (commented):
Thank you - clearly this was much more work than we thought.

Have added some inline comments and lots of thoughts on naming - sorry!

The fact this change needed changes accross 10+ files is a code smell - our existing code is not open to extension, closed to modification, and not following the SRP as the logic is all over the place. We pass around lots of file paths instead of layer objects. Ideally this change would need just changes to the Layer object, and whever is responsible for saving/loading it. Lets all try and bear that in mind for future work.

But for now maybe a bit more work on it? Think that at minimum some of the logic should get pushed into layer objects.

I dont think we should support 2 types of json files with inline vectors and external vectors in csv files - just adds complexity. We could retire the inline ones? 

Seeing as we already have valuetype we can use that, and get rid of that filepath in the metadata section which is a bit hidden for my taste. Something like this using $ref which is in some of the json specs?

```json
"value_type": "vector_csv_file",
"values": {
    "$ref": "./some_data_file_name.csv"
    // could have other meta data here if we needed it
 }
```

src/imap_mag/io/file/CalibrationDataPathHandler.py
```diff
@@ -0,0 +1,22 @@
...
+
+@dataclass
+class CalibrationDataPathHandler(CalibrationLayerPathHandler):
+    """
+    Path handler for calibration layer science data.
+    Designed to handle the special internal case of calibration layer science data.
+    """
+
+    @property
+    def extra_descriptor(self) -> str:
```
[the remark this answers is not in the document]
mhairifin: Not sure I agree with this, though I think being colocated in file systems would be good I think having two files with the same name and completely different content is confusing and unclear. Willing to be convinced though!

tests/datastore/calibration/layers/2025/10/imap_mag_noop-layer-data_20251017_v001.csv
mhairifin: In general I have been calling it time when it is a time/timestamp so that epoch can be used when it is actually an epoch (ns since J2000), as it feels clearer. I'm willing to change this though if it would be simpler to just call it epoch everywhere.

src/imap_mag/cli/calibrate.py
```diff
@@ -60,18 +63,25 @@ def gradiometry(
...
+    calibration_metadata_handler = CalibrationMetadataPathHandler(
+        calibration_descriptor=method.short_name, content_date=date
+    )
+    calibration_data_handler = CalibrationDataPathHandler(
         calibration_descriptor=method.short_name, content_date=date
     )
-    result: Path = calibrator.run_calibration(
-        work_folder / Path(calibration_layer_handler.get_filename()),
+
+    metadata_path, data_path = calibrator.run_calibration(
```
mhairifin, line 73: I think that rather than passing run_calibration two things, it should just get one metadata handler or one thing - in general I think we should try and avoid the double files having to be passed around everywhere, and I think it should be more opaque that in the background two files are being created

src/imap_mag/io/file/CalibrationDataPathHandler.py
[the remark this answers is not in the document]
alastairtree: Yeah maybe there are no good answers to this! I will keep trying. How about? 
```
imap_mag_noop-layer_20250814_v001.json
imap_mag_noop-layer_20250814_v001_vectors.csv
```
Thoughts?

src/mag_toolkit/calibration/Layer.py
```diff
@@ -22,11 +25,32 @@ class Layer(BaseModel, ABC):
...
-    def from_file(cls, path: Path):
+    def from_file(cls: type[T], path: Path) -> T:
         with open(path) as fid:
             as_dict = yaml.safe_load(fid)
         model = cls(**as_dict)
-        return model
+
+        # If data is defined in a separate file load it, otherwise
+        # leave it as is.
+        if not model.metadata.data_filename:
```
mhairifin, line 35: I don't think we should account for this anymore - I don't think we need to support the old layer files, I think instead we should entirely support the new layer file format

tests/datastore/calibration/layers/2025/10/imap_mag_noop-layer-data_20251017_v001.csv
alastairtree: Agree with your naming convention. I meant which should be stored in the layer file, i would have assumed epoch but i guess there was a good reason to use time?

src/mag_toolkit/calibration/ScienceLayer.py
```diff
@@ -114,13 +115,25 @@ def from_file(cls, path: Path):
         else:
             return super().from_file(path)
 
+    @classmethod
+    def _load_data_file(cls, path: Path, existing_model) -> "ScienceLayer":
```
mhairifin, line 119: This load_data method is I think exactly the same for `ScienceLayer` and `CalibrationLayer` and could be abstracted up to `Layer` (aside from the error message, but I think that's okay)

tests/datastore/calibration/layers/2025/10/imap_mag_noop-layer-data_20251017_v001.csv
mhairifin: Ah sorry - I have gone with time in the spirit of the layer files being human readable and easy to plot without having to perform any conversion. If we want to use epoch I think that's fine, but would probably like to write a human readable time as well, so that you could in theory physically open the file and understand immediately what timestamps it contained vectors for.

src/imap_mag/cli/apply.py
```diff
@@ -29,29 +31,85 @@ class FileType(Enum):
...
             fetch_file_for_work(
-                versioned_cal_file, appSettings.work_folder, throw_if_not_found=True
+                versioned_metadata_file,
+                appSettings.work_folder,
+                throw_if_not_found=True,
+            )
+        )
+
+        # Get data file
+        with open(versioned_metadata_file) as fid:
```
mhairifin, line 64: Not sure I understand why this is being opened with yaml, as opposed to using `CalibrationLayer` to open the file and verify its contents. Think this is duplicating a bit of verification work that pydantic could handle for you.

src/imap_mag/io/file/CalibrationDataPathHandler.py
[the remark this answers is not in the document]
mhairifin: I think I would prefer `data` or `offsets` as opposed to vectors because although they are vectors, we are usually using `vectors` to indicate measured magnetic field vectors which these are not

mhairifin:
@mfacchinelli Yes, you are correct that a science layer needs to be able to be loaded from a CDF file where the data will be colocated with the metadata. And yes, value type is currently used to specify what kind of `values` are stored - e.g. vector by vector values, or values that should be interpolated across vectors. We may in future also need to support rotation through these values. Agree on unfortunate issues with using $ref with matlab

src/imap_mag/io/file/CalibrationDataPathHandler.py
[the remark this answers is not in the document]
mfacchinelli: Agreed on:
```
imap_mag_noop-layer_20250814_v001.json
imap_mag_noop-layer-data_20250814_v001.csv
```

src/imap_mag/cli/apply.py
```diff
@@ -29,29 +31,85 @@ class FileType(Enum):
...
+            as_dict = yaml.safe_load(fid)
+
+        if not (("metadata" in as_dict) and ("data_filename" in as_dict["metadata"])):
+            logger.error(
+                f"Metadata file {versioned_metadata_file} must have the correct format: 'metadata' or 'metadata.data_filename' fields are missing."
```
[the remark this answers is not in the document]
mfacchinelli: As discussed, `$ref` cannot be used. I think `data_filename` makes sense, given the discussion on how `vector` is too "magnetic field" adjecent.

src/mag_toolkit/calibration/ScienceLayer.py
mfacchinelli: Cannot be a member of the superclass because it needs to reference the `values` property, which is defined in the subclasses

src/mag_toolkit/calibration/Layer.py
mfacchinelli: I thought that some layers can be defined in CDF, and thus will not have a separate `data_filename`

PR #134, opened 2025-08-15
feat: WEB PODA service ID (flight/preflight) is configurable

alastairtree:
Now the service ID comes from config in the URL base from ENV var MAG_FETCH_BINARY_API_URL_BASE which is now part of the platform deploy

### DIFF
--- a/src/imap_mag/client/WebPODA.py
+++ b/src/imap_mag/client/WebPODA.py
@@ -118,13 +118,14 @@ class WebPODA:
         end_value: str = end_date.strftime("%Y-%m-%dT%H:%M:%S")
         time_var = "ert" if ert else "time"
 
-        url = (
+        # default to the pre-launch system ID if one is not passed in the URL
+        url_base: str = (
             f"{urllib.parse.urljoin(self.__webpoda_url, 'packets/SID2/')}"
-            f"{packet}.{extension}?"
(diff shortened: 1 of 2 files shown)


src/imap_mag/client/WebPODA.py
```diff
@@ -118,13 +118,14 @@ def __download_from_webpoda(
...
             f"{urllib.parse.urljoin(self.__webpoda_url, 'packets/SID2/')}"
-            f"{packet}.{extension}?"
-            f"{time_var}%3E={start_value}&"
-            f"{time_var}%3C{end_value}&"
-            f"{data}"
+            if "packets/SID" not in self.__webpoda_url
+            else self.__webpoda_url
         )
+        url_base = url_base.rstrip("/")
+        url = f"{url_base}/{packet}.{extension}?{time_var}%3E={start_value}&{time_var}%3C{end_value}&{data}"
```
mfacchinelli, line 128: Can we keep the previous formatting? It was easier to read all on separate lines

mfacchinelli (approved):
Thanks!

src/imap_mag/client/WebPODA.py
alastairtree: Ha - i changed this because I felt the exaact opposit! I have had another go at improving readaility

PR #144, opened 2025-09-02
fix: Use MATLAB tag v0.2.0

mfacchinelli:


alastairtree:
Has no meaningful change, just tidying up. Thanks!

PR #150, opened 2025-09-03
feat: add L2 files and offset files to DB

mhairifin:
Adds the L2 files and the offset files to the database

### DIFF
--- a/src/imap_db/model.py
+++ b/src/imap_db/model.py
@@ -26,7 +26,7 @@ class File(Base):
     version: Mapped[int] = mapped_column(Integer())
     hash: Mapped[str] = mapped_column(String(64))
     size: Mapped[int] = mapped_column(Integer())
-    content_date: Mapped[datetime] = mapped_column(DateTime())
+    content_date: Mapped[datetime] = mapped_column(DateTime(), nullable=True)
     creation_date: Mapped[datetime] = mapped_column(
         DateTime(), server_default=func.now()
     )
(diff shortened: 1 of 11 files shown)


src/imap_db/model.py
```diff
@@ -26,7 +26,7 @@ class File(Base):
     version: Mapped[int] = mapped_column(Integer())
     hash: Mapped[str] = mapped_column(String(64))
     size: Mapped[int] = mapped_column(Integer())
-    content_date: Mapped[datetime] = mapped_column(DateTime())
+    content_date: Mapped[datetime] = mapped_column(DateTime(), nullable=True)
```
alastairtree, line 29: Sorry - if you want to change the database schema you need to do a database migration using alembic. The prod database needs to have to schema changed to make that field nullable. The CLI tools should make one automatically but you need to ask for it. Let me know if you want me to show you this,

src/imap_mag/io/file/IFilePathHandler.py
```diff
@@ -28,6 +29,11 @@ def get_full_path(self, parent: Path = Path()) -> Path:
         """Get the full path of the file."""
         return parent / self.get_folder_structure() / self.get_filename()
 
+    @abc.abstractmethod
+    def get_date_for_indexing(self) -> datetime | None:
```
alastairtree, line 33: Minor but think I prefer `get_content_date_for_indexing()`

alastairtree (changes requested):
Nice, but does need a migration as will fail once deployed to a real db

PR #194, opened 2025-11-07
task: Add science to I-ALiRT quicklook

mfacchinelli:
Tim asked to add science to the I-ALiRT quicklook, when it's fixed. We are not there yet, but this can be merged when it is.

### DIFF
--- a/src/imap_mag/plot/plot_ialirt_files.py
+++ b/src/imap_mag/plot/plot_ialirt_files.py
@@ -98,6 +98,8 @@ def create_figure(
             "mag_hk_hkp8v5c_danger",
         ]
     ):
+        # Multiply by (1 + i * 0.1) to offset the lines for better visibility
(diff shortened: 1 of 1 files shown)


src/imap_mag/plot/plot_ialirt_files.py
```diff
@@ -61,9 +61,31 @@ def create_figure(
...
+    # GSE science
     ax00 = fig.add_subplot(gs[0, 0:2])
 
+    for i, b in enumerate(
+        [
+            "mag_B_GSE_x",
+            "mag_B_GSE_y",
+            "mag_B_GSE_z",
+            "mag_B_magnitude",
+        ]
```
alastairtree: Have you asked Tim what he wants plotted? For the I-ALiRT website they just plot Bz and |B| so worth checking, and all on one plot will be hard to view

PR #240, opened 2026-03-17
Add file analysis flow for finding gaps and bad files

alastairtree:
- New flow will analyse all files in the datastore and tries to be general purpose
- If they are csv or cdf it looks for gaps in the timestamps
- configurable based on file pattern
- Saves results into the database. We can the replicate this to Azure and use grafana to make a dashboard of the results


alastairtree:
@mhairifin I would not be surprised if this is a bit buggy, and the config is probably in need of more thought but I reckon this is already useful enough to be able to start making a gaps dashboard. Even just plotting the number of records in every file would be helpful and this does a fair bit more than that?

src/imap_mag/config/FileAnalysisConfig.py
```diff
@@ -0,0 +1,30 @@
...
+    columns_to_check: list[ColumnCheckConfig] = []
+
+
+class FileAnalysisConfig(BaseModel):
+    """Configuration for the file index flow."""
+
+    paths_to_match: list[str] = []  # fnmatch patterns for finding files to index
+    file_patterns: list[FileAnalysisPatternConfig] = []  # per-file pattern configs
+    default_gap_threshold_seconds: int = 60  # 60 seconds = 1 minute
+    nan_sentinel: float = -9.999999848243207e30
```
mhairifin, line 30: I think this is the nan value for epochs - but for CDF_FLOAT values it should be -1e31 (I think - double check)

src/imap_mag/data_pipelines/AnalyseFilesStage.py
```diff
@@ -0,0 +1,772 @@
...
+            first_timestamp = None
+            last_timestamp = None
+
+            if datetime_col:
+                try:
+                    epoch_data = cdf.varget(datetime_col)
+                    if epoch_data is not None and len(epoch_data) > 0:
+                        record_count = len(epoch_data)
+                        # Convert CDF epoch to datetime
+                        times = cdflib.cdfepoch.to_datetime(epoch_data)
```
mhairifin, line 143: Functionality here looks fine - but logistically loading in burst mode files might be time intensive (loading one is fine, loading a lot might take time - and sometimes time conversion on large numbers of datetimes takes time as well). I think we want to do it anyway, but it might be worth thinking about putting the file analysis jobs in a separate queue or running them at a quiet time - I am concerned about scenarios where we download a lot of l1d-burst and it could occupy the pipeline when we might be wanting to prioritise other tasks. I might be overestimating the time it takes to do the analysis though!

src/prefect_server/fileAnalyserFlow.py
```diff
@@ -0,0 +1,97 @@
...
+        return IndexByDateRangeRunParameters(
+            modified_after=modified_after,
+            modified_before=modified_before,
+            progress_mode=ProgressUpdateMode.NEVER_UPDATE_PROGRESS,
+        )
+    return AutomaticRunParameters()
+
+
+@flow(name=PREFECT_CONSTANTS.FLOW_NAMES.FILE_ANALYSER, log_prints=True)
+async def file_analyser_flow(
```
mhairifin, line 55: Similar to my comment above about limited time and computation - should this have a configurable maximum number of files (if it is something that could take up time on the pipeline)

tests/unit/test_AnalyseFileStage.py
```diff
@@ -0,0 +1,819 @@
+"""Unit tests for AnalyseFilesStage."""
```
mhairifin, line 1: Don't think there is a test here for a file with no data in it? Is that possible and how would the file analysis handle it?

src/imap_mag/data_pipelines/AnalyseFilesStage.py
```diff
@@ -0,0 +1,772 @@
...
+
+            # Find the datetime column
+            datetime_col = self._find_datetime_column_cdf(variables, pattern_config)
+
+            timestamps = None
+            record_count = None
+            first_timestamp = None
+            last_timestamp = None
+
+            if datetime_col:
```
mhairifin, line 137: This function is quite long with a few distinct steps - worth refactoring to make it slightly more readable?

src/imap_mag/data_pipelines/AnalyseFilesStage.py
```diff
@@ -0,0 +1,772 @@
...
+        datetime_col = self._find_datetime_column_csv(
+            df.columns.tolist(), pattern_config
+        )
+
+        timestamps = None
+        first_timestamp = None
+        last_timestamp = None
+
+        if datetime_col:
+            try:
```
mhairifin, line 327: Same comment here on long function - also with a lot of overlap with the index_cdf function - any way of sharing functionality between them, or too difficult because of the different data formats?

mhairifin (commented):
Overall looks good - main concern is a) I am not sure how this would handle an empty file, and b) I am aware from past usage of the pipeline that sometimes things take an unexpectedly long time, and the pipeline is not always managing priority well. Do you have a sense of how long a FileAnalysis job takes for a standard burst file? I think it could be sensible to put in eg a maximum number of files it will analyze at a time, or even separate analysis jobs when we deploy for each type of file, so we have good ways of managing the capacity of the pipeline and allowing it to interleave these jobs in at quiet times. What do you think?

PR #258, opened 2026-04-09
fix: Do not crash when applying an empty quality layer

alastairtree:
Following onfrom the previous pr #256 this now applies that empty layer ratehr than errors


alastairtree:
@mhairifin I think this addresses a bunch of your code review comments and fixes some bugs and edge cases in the quality layres so hat now we can use them over a whole month

src/mag_toolkit/calibration/calibrators/SetQualityAndNaNCalibration.py
```diff
@@ -74,45 +74,97 @@ def run_calibration(
...
+        # create boundary change rows for the ends of all time windows defined in the config file to reset values back to their previous values.
+        # If the window ends after the end of the day then there will be zero end_rows because we are reading from end_filtered
+        end_rows = pd.DataFrame(
+            {
+                CONSTANTS.CSV_VARS.EPOCH: end_filtered["window_end"].values,
+                CONSTANTS.CSV_VARS.OFFSET_X: np.zeros(len(end_filtered)),
+                CONSTANTS.CSV_VARS.OFFSET_Y: np.zeros(len(end_filtered)),
+                CONSTANTS.CSV_VARS.OFFSET_Z: np.zeros(len(end_filtered)),
+                CONSTANTS.CSV_VARS.TIMEDELTA: np.zeros(len(end_filtered)),
+                CONSTANTS.CSV_VARS.QUALITY_FLAG: [pd.NA] * len(end_filtered),
```
mhairifin: should quality flag and quality bitmask be NA here? should it not just be zero?

src/mag_toolkit/calibration/CalibrationApplicator.py
```diff
@@ -300,6 +303,25 @@ def _expand_boundary_changes_to_every_epoch(
...
         science_index = pd.DatetimeIndex(science_epochs)
+
+        if change_points.empty:
+            zero_df = pd.DataFrame(
+                {
+                    CONSTANTS.CSV_VARS.OFFSET_X: 0.0,
+                    CONSTANTS.CSV_VARS.OFFSET_Y: 0.0,
+                    CONSTANTS.CSV_VARS.OFFSET_Z: 0.0,
+                    CONSTANTS.CSV_VARS.TIMEDELTA: 0.0,
+                    CONSTANTS.CSV_VARS.QUALITY_FLAG: pd.NA,
```
mhairifin: Same note here - not sure I understand why these are N/A rather than zero

mhairifin (commented):
I don't think I understand the use of NA here and why it is being used instead of zero as a quality flag/bitmask, especially as it gets filled as zero. It's obviously being used very intentionally so presumably it is fixing a bug/edge case I'm not seeing!

alastairtree:
Yeah thats fair - not really explained what I am trying to address sorry.

If the CSV defines a period in which we set the flag/mask, we should not (i think) be setting all other periods to 0. Also now we can have a empty quality layer for a non-flagged day this is more of an issue - your matlab code may flag times and those should not be overridden by the quality layer back to being zeros. That is what we are fixing, the current version of code sets all periods outside of the flagged times to be zero whereas now the NA marks the times/boundaries where we want the lower layer to define the flag/mask and not default to zero.

This then allows us to define in the config file if we want times to be flagged false/0 or if we want to let an earlier layer proagate through.

So this change means that now if I have for 16th Jan
- Offset layer with all day flag=0 mask=0 except 4AM-5AM bitmask=2
- Quality config layer 2AM-3AM to be flag=1 bitmask=3 

Then the quality layer only overrides 2-3AM and the rest of the day flows through from the offset layer meaning 4-5AM will be bitmask=2 and not reset to 0. Also 17th Jan is not all zero, and instead is just a passthrough layer.

Make any sense?

src/mag_toolkit/calibration/calibrators/SetQualityAndNaNCalibration.py
alastairtree: NA shows that we not setting the flag to zero for the rest of the day , we are setting it to whatever the flag is in the previous layer. In this case if a MATLAB layer flagged a time then it can still flow through the layers if it is after window_end, and not just be overridden to zero

src/mag_toolkit/calibration/CalibrationApplicator.py
alastairtree: Say config file has one row, 16th jan 2AM-3AM as flag=1, but now we process the 17th Jan.

In this case we have no times in the current day that match the config file. We do not want to set all quality flags to zero, we want the flags to passthrough from the manual layer.

If we did want to set all day to flag=0 we could do that explicitly in the config file

mhairifin:
> Make any sense?

I'm not sure - since it is a layer the quality should never `override` the pre-existing data, the quality part of the layers should always be combined by a bitwise or, so that all the data is carried through - so e.g. if this quality layer sets a flag to 0, it should not overrride anything in the offsets layer that has set the quality flag to 1 (for example, if the pivot platform is moving) - they should just be combined by the bitwise or. An all zero quality layer should have no impact if everything is being combined correctly I think?

alastairtree:
> An all zero quality layer should have no impact if everything is being combined correctly I think?

Agreed - but without this PR the corrent code is setting times outside the configured windows to zero and not bitwise or'ing them

> the quality part of the layers should always be combined by a bitwise or, so that all the data is carried through

But then you cannot deliberatly reset/clear the flag/mask. If the config file for the top layer says 2AM-3AM flag=0, mask=0 I would assume they would be zero at that time and whatever value is in the lower layers?

alastairtree:
I have updated this PR so that 0 for quality flag or mask in a layer file is effectively a no-op since x OR 0 = x. 
Now in the config csv you can pass in "-1" to clear a 1, or "-5" to clear bits 1 (=1) and 3 (=4) or -65536 to clear all mask bits and force all bits to zero. Also added some comments to give us some docs on CSV contents to explain this. Hopefully we can merge this today.

mhairifin (approved):
Looks good to me, the negative numbers is a good solution to the issue I think!

PR #260, opened 2026-04-10
feat: only generate L2 frames required

alastairtree:
We have already updated imap_processing so now just generate the l2-pre frames we need and not all of them

### DIFF
--- a/src/mag_toolkit/calibration/CalibrationApplicator.py
+++ b/src/mag_toolkit/calibration/CalibrationApplicator.py
@@ -10,6 +10,7 @@ import pandas
 import spiceypy
 from cdflib.xarray import cdf_to_xarray, xarray_to_cdf
 from imap_processing.mag.l2 import mag_l2, mag_l2_data
+from imap_processing.mag.l2.mag_l2_data import ValidFrames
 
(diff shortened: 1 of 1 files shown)


src/mag_toolkit/calibration/CalibrationApplicator.py
```diff
@@ -281,6 +283,19 @@ def apply(
 
         return (files_created, created_offsets_filepath)
 
+    @staticmethod
+    def _get_l2_frames(reference_frames: list[ReferenceFrame]) -> list[ValidFrames]:
+        frame_mapping = {
```
mhairifin, line 288: Looks good - only question is if this is the most sensible place for this mapping - for example it might make more sense for the mapping to be integrated into the `ReferenceFrame` class, so a reference frame returns its own ValidFrame, which might be the place one would expect to find it

mhairifin (approved):
Looks good! One small comment, but it's very minor so happy to just go with your judgment.

src/mag_toolkit/calibration/CalibrationApplicator.py
alastairtree: I left it in here deliberately because this is the only file in the whole project that depends on the imap_processing codebase

PR #262, opened 2026-04-15
fix: Resolve files by name & layers versioned together

alastairtree:
- Ensure that all places we pass filenames (science, layer,spice mk) in cal process cli and prefect ui accept both abs paths, datastore relative paths and just file names which are resolved dynamically
- Ensure layer json and csv files never have a different version number

### DIFF
--- a/src/imap_mag/io/FileFinder.py
+++ b/src/imap_mag/io/FileFinder.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Literal, overload
 
+from imap_mag.db.Database import Database
 from imap_mag.io.file import (
     CalibrationLayerPathHandler,
     IFilePathHandler,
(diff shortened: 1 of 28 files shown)


src/imap_mag/io/FileFinder.py
```diff
@@ -195,26 +206,25 @@ def _keep_highest_version_layers_only(filenames: list[str]) -> list[str]:
                 best[key] = (name, handler.version)
         return [name for name, _ in best.values()]
 
-    def find_science_file(
-        self, date: datetime, mode: ScienceMode, sensor: MAGSensor
+    def find_latests_science_by_date(
```
mhairifin:
```suggestion
    def find_latest_science_by_date(
```

src/imap_mag/io/FileFinder.py
```diff
@@ -292,3 +302,105 @@ def __find_files_and_sequences(
...
+            else file_name_or_path
+        )
+
+        # path already exists → return immediately
+        if path.exists():
+            logger.debug("Resolved '%s' via existing path.", file_name_or_path)
+            return path
+
+        # Relative path from data_store
+        candidate = self._data_store / path
```
mhairifin: Only comment is ordering here - I think maybe I would expect a file explicitly placed into the work folder to have precedence over one in the datastore

mhairifin (approved):
Looks good! couple small comments

src/imap_mag/io/FileFinder.py
alastairtree: Agreed!

PR #263, opened 2026-04-15
fix: Ensure ialirt errors when zero checks are performed

alastairtree:
- Fix mag_hk_mode forbidden check: CSV now emits string values ("Safe", "Debug") after the poll ialirt HK schema change, replacing the old integer mapping values [2, 4] that never fired
- Raise RuntimeError in check_ialirt_files when no validation rules matched any CSV column, preventing silent no-op runs
- trigger the check flow with a hk file not a science file as the science files cannot be checked

### DIFF
--- a/src/imap_mag/packet_def/ialirt_4.05.yaml
+++ b/src/imap_mag/packet_def/ialirt_4.05.yaml
@@ -18,8 +18,8 @@ ialirt_human_readable_names:
   - mag_hk_multbit_errs: Multi-bit Errors
   - mag_hk_fib_saturated: FIB Saturation
   - mag_hk_fob_saturated: FOB Saturation
-  - mag_hk_fib_isvalid: FIB Status
-  - mag_hk_fob_isvalid: FOB Status
+  - mag_hk_pri_isvalid: FIB Status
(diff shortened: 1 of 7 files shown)


src/imap_mag/packet_def/ialirt_4.05.yaml
```diff
@@ -18,8 +18,8 @@ ialirt_human_readable_names:
   - mag_hk_multbit_errs: Multi-bit Errors
   - mag_hk_fib_saturated: FIB Saturation
   - mag_hk_fob_saturated: FOB Saturation
-  - mag_hk_fib_isvalid: FIB Status
-  - mag_hk_fob_isvalid: FOB Status
+  - mag_hk_pri_isvalid: FIB Status
+  - mag_hk_sec_isvalid: FOB Status
```
mhairifin:
```suggestion
  - mag_hk_pri_isvalid: FOB Status
  - mag_hk_sec_isvalid: FIB Status
```

alastairtree: Good spot!

PR #264, opened 2026-04-17
Feat/download spin tables

alastairtree:
Stacked PR! - review and Merge this after the #258 

CLI and flow for spin tables


mhairifin (approved):
Looks good! Will be great to have this automatically downloaded as I think it's our last local dependency :)

PR #265, opened 2026-04-20
fix: layer data file version is locked once chosen even when identical file already exists

alastairtree:
When running SetQualityAndNaN calibration twice with SaveMode.LocalAndDatabase, the second run produced identical data CSV content. DBIndexedDatastoreFileManager found a hash match at v001 in the DB and redirected the handler back to v001, while the layer JSON had already been written referencing data_filename=v002.csv. Fixed by allowing layer files version to be locked.

Bumps version to 5.1.1.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 requires-python = ">=3.11,<3.15"
 name = "imap-mag"
-version = "5.1.0"
+version = "5.1.1"
 description = "Process IMAP data"
 authors = [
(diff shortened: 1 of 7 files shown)


mhairifin (approved):
It seems slightly unintuitive to me that the second run would create a `v002` file, rather than when it realises the hash is the same not just leaving it at v001, and waiting to write the data file until it is confident it is not a repeat. However, I think this is also a viable solution, and the quality layers are small so don't think this will cause us any practical storage issues.

PR #268, opened 2026-04-28
feature: Layer files are deduplicated on creation based on the science data

alastairtree:
…librate with database

When running SetQualityAndNaN calibration twice with SaveMode.LocalAndDatabase,
the second run produced identical data CSV content. DBIndexedDatastoreFileManager
found a hash match at v001 in the DB and redirected the handler back to v001,
while the layer JSON had already been written referencing data_filename=v002.csv.
Fixed by allowing layer files version to be locked.

Bumps version to 5.1.1.


src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -93,13 +96,106 @@ def from_filename(
...
         super().increase_sequence()
-        if hasattr(self, "_version_change_callback"):
-            self._version_change_callback(self)
         logger.debug(
             f"Increased version to {self.version} for file {self.get_filename()}."
         )
+
+    # ── Content-identity overrides for JSON metadata files ──────────────────
+
+    def _companion_csv_path(self, alongside: Path) -> Path:
```
alastairtree, line 102: alongside should be renamed layer_json_file

alastairtree:
@mhairifin this is a second attempt to replace #265

alastairtree:
I have now tested this and it migrates all the layer files to the new format and they are now deduped on save as expected. @mhairifin I am merging but please do a review and i will pick up and fixes later

PR #270, opened 2026-04-28
Update dependencies and fix data filtering in FetchIALiRT

alastairtree:
- Updated the 'prefect[shell]' dependency version from 3.6.20 to 3.6.28 in pyproject.toml.
- Modified the data filtering logic in FetchIALiRT to ensure proper comparison of datetime index values with min_daily_date.


alastairtree:
Deployed to test and have checked the ialirt flows, and a couple of other selected flows and seems ok so closing as done

PR #277, opened 2026-05-01
feature: add IMAP-Hi 45 and Hi 90 ESA STEP HK download flows

alastairtree:
Adds two new Prefect flows (poll-hi45-esa-step, poll-hi90-esa-step)
that download daily ESA STEP housekeeping CSV files from the WebTCAD
LaTiS API for TMID 58238 (Hi-45) and 58309 (Hi-90), modelled on the
existing poll_lo_pivot_platform_flow. Introduces a generic
DownloadWebTCADCsvFilesStage and parameterised HiEsaStepPipeline so
the same shape can serve any HKWebTCADItems entry, plus matching
imap-mag fetch imap-hi45-step / imap-hi90-step CLI entrypoints and
WireMock-based tests for both date-range and automatic progress modes.

@mhairifin this is untested but i think it is 95% there


mhairifin (approved):
Apologies for the late review - this all looks good to me, and like a sensible refactoring of the lo pivot platform code to be a bit more general for other HK

PR #284, opened 2026-05-03
fix: pool SQLAlchemy connections to prevent exhausting postgres

### LINKED ISSUE #283: Leaking SQL Connections
When running lots of flows, many calibrate + apply for example for many days we get errors about no more connections on the database. Suggests we are not pooling database connections correctly - should be one per flow container.

Also seeing tests in parallel getting similar SQL errors.

alastairtree:
Each Database() call was creating a fresh engine with its own connection pool, so N concurrent flows/tests consumed N*pool_size connections and hit the postgres 'too many clients' limit.

Two fixes:
1. Cache engines at class level keyed by URL so all Database instances with the same URL share one pool.
2. Wrap the six bare self.session().execute() read methods in explicit 'with self.session() as session:' blocks so connections are returned to pool immediately rather than relying on garbage collection.

Regression tests added: test_multiple_database_instances_share_engine (would fail before fix) and test_database_read_methods_return_connections_to_pool.

### DIFF
--- a/src/imap_mag/db/Database.py
+++ b/src/imap_mag/db/Database.py
@@ -1,9 +1,12 @@
 import functools
 import logging
 import os
+import threading
 from datetime import datetime
+from typing import ClassVar
 
 from sqlalchemy import create_engine, or_, select
+from sqlalchemy.engine import Engine, make_url
 from sqlalchemy.orm import Session, sessionmaker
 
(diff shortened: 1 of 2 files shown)


mhairifin (approved):
Looks good, this all makes sense

PR #291, opened 2026-06-01
Create a pipeline to azure for i-alirt hit swe swapi and codice data

### LINKED ISSUE #271: Create a pipeline to azure for I-ALiRT HIT, SWE, SWAPI and CODICE data
Add a data pipeline to this repo that downloads the latest HIT, SWE, SWAPI and CODICE data from the I-ALiRT APIs into CSV data files, one per day and save them in the datastore.

We already have data pipelines for MAG science and MAG HK already but they are going to need to change for the new pipelines - see this code for the existing [prefect data flows ](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/prefect_server/pollIALiRT.py)

We have newer framework/infrastructure code for pipeline building which allows us to more easily build pipelines, track progress and reuse shared code in Stages so that every pipeline can reuse similar logic. For example code to put to [PublishFileToDatastoreStage](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/data_pipelines/PublishFileToDatastoreStage.py), or a stage to [update the workflow tracking progress table](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/data_pipelines/SaveProcessingDatesStage.py). An example of a pipeline that uses the pipeline framework is at https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/prefect_server/pollLoPivotPlatform.py and that calls the [LoPivotPlatformPipeline ](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/data_pipelines/LoPivotPlatformPipeline.py#L22) class
I would be helpful for reusability if the new IMAP IALIRT pipelines are built using the new Pipeline/Stages framework.

The pipelines can all also be executed with a simple CLI so that the prefect flow code is really just a scheduler and wrapper around plain CLI code that is framework agnostic. For example the MAG IALIRT CLI command is in https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/cli/fetch/ialirt.py and can be called like so: `imap-mag fetch ialirt --start-date 2025-01-02 --end-date 2025-01-03` 
The CLI needs to be maintained and extended for the new instruments. I suggest you add a new instrument argument that defaults to mag, e.g. `imap-mag fetch ialirt --instrument swe --start-date 2025-01-02 --end-date 2025-01-03` or `imap-mag fetch ialirt --instrument hit --start-date 2025-01-02 --end-date 2025-01-03`

To get the instrument data you can reuse the existing API client - see https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/client/IALiRTApiClient.py. This calls a package issued by the mission which has documentation at https://pypi.org/project/ialirt-data-access/
The list of instrument data products we want to download are
- hit
- mag (already done)
- codice_lo
- codice_hi
- swapi
- swe

Example API cURLs:
```
curl "https://ialirt.imap-mission.com/space-weather?instrument=hit&time_utc_start=2026-04-22T05:30:00&time_utc_end=2026-04-22T08:30:00"
curl "https://ialirt.imap-mission.com/space-weather?instrument=swe&time_utc_start=2026-04-22T05:30:00&time_utc_end=2026-04-22T08:30:00"
```

Each instrument json data should be converted and saved in a CSV file per day. The current MAG data is in the datastore in files arranged by year and month, so 1st Feb 2026 is in
`[datastore root]\ialirt\2026\02\imap_ialirt_20260201.csv`. Unfortunately the MAG data does not have MAG in the name - oops - but lets leave that be. I suggest you flow instrument name into the files so you get `[datastore root]\ialirt\2026\02\imap_ialirt_hit_20260201.csv`, `[datastore root]\ialirt\2026\02\imap_ialirt_swe_20260201.csv` etc in the same folder

When downloading we use the Workflow Progress database table with a key like MAG_IALIRT or MAG_IALIRT_HK to keep track of where we have already downloaded up to and then crawl forwards in time to only download the new data. You should track each IALIRT instrument separately with their own key and crawl each one in turn.
We also have flow arguments allowing us to manually catch up/re-download earlier data by passing date args, but the default is to just download the new stuff. When downloading new data the rows get appended to the existing CSV files if it already exists so that we do not re-download the same data for today every 5minutes. This code is in https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/download/FetchIALiRT.py

In prod, the 2 current I-ALiRT flows (mag, mag_hk) run (in containers) in prefect once per hour, and poll for data every (configurable) 5 minutes. They automatically shut down just before the hour boundary. We don't really want 7 instruments plus 1 HK (so 8!) containers running all the time - that is a waste of resources. Instead refactor to have one I-ALiRT flow that runs hourly. Within that one [Prefect Flow](https://docs.prefect.io/v3/concepts/flows) I suggest you have a main loop that runs every (configurable) 5 minutes and which launches one [Prefect Task
](https://docs.prefect.io/v3/concepts/tasks) to get the latests data for per IALIRT instrument in turn (so 8 tasks), including the mag_hk, and mag. Then once they have all downloaded sleep for a while and start again launching another 8 tasks. This setup will keep logs separate, make observability easy and not waste resources.

For the data to appear in the database we can reuse [the existing postgres data pipeline](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/prefect_server/postgresUploadFlow.py#L92) to read CSV files and map them to a corresponding postgres table using [my library crump](https://alastairtree.github.io/crump/). All you need to do is add to the config files so it knows how to find the files and how to map them:
1. Add the path e.g. `ialirt/**/imap_ialirt_swe_*.csv` to the new CSV files to the main app config file in 
https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/imap-mag-config.yaml#L107
```
postgres_upload:
    crump_config_path: imap-db-ingest-config.yaml
    database_url_env_var_or_block_name: 'imap-database'
    enable_history: true
    paths_to_match:
       .........
        - "ialirt_hk/**/imap_ialirt_hk*.csv"
        - "ialirt/**/imap_ialirt_*.csv"
         - "ialirt/**/imap_ialirt_swe_*.csv"
```
2. Add a new table def like the IALIRT MAG csv data mapping is at https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/imap-db-ingest-config.yaml#L11028
You can generate the crump config file  from the csv data file using the CLI easily once you have CSV file you want to ingest, e.g. `crump
 prepare imap_ialirt_swe_20260201.csv --config imap-db-ingest-config.yaml --job ialirt_swe` will spit out the required config to ingest that data. you might then need to tweak the config and check the data types etc.

miruuna:
Added ability to fetch data for any iALiRT instrument. 

Main changes:
- CLI commands can now take the instrument as an argument. HK data for specific instruments can be requested using `<instrument_name)_hk`. E.g. `imap-mag fetch ialirt --instrument hit --start-date 2025-01-02 --end-date 2025-01-03`. Valid instruments must be added to CONSTANTS in `imap_mag.util.constants`
- Added a new pipeline stage  (`DownloadIALiRTStage.py`))
- Added new IALirt data pipeline (`IALiRTStage.py`)
- Updated Prefect workflow (`pollIALiRT.py`) - - A polling flow runs every hour which runs tasks to download instrument data (both hk and science) every 5 minutes:
  - Added a task for polling iALiRT data (`run_ialirt_polling_pipeline_task()`) - runs every 5 minutes. It polls both science and HK data
  - Updated polling flow (`poll_ialirt_flow()`)
  - Added migration to change records in the database and legacy mag filenames


miruuna:
@alastairtree here is just a draft PR for the implementation, whenever you get a chance to check if I am on the right track. Most tests fail locally as I do not have the API `auth_code`.  The CLI command to download the data works locally if using the non-dev API `url_base` (`https://ialirt.imap-mission.com`).  I am not sure if the data from the other instruments need any more processing, as it's being done for mag and mag_hk.

src/prefect_server/pollIALiRT.py
```diff
@@ -118,6 +153,8 @@ async def poll_ialirt_flow(
             }
         ),
     ] = True,
+    start_date=None,
+    end_date=None,
```
alastairtree: These dates are provided by FetchByDatesRunParameters?

imap-db-ingest-config.yaml
```diff
@@ -11109,3 +11108,256 @@ jobs:
...
+    filename_match: imap_ialirt_*.csv
+  ialirt_swe:
+    target_table: ialirt_swe
+    id_mapping:
+      time_utc:
+        db_column: id
+        type: datetime
+        nullable: false
+    columns:
+      instrument:
```
alastairtree: Suggest you optimise the db table a bit to reduce junk
- delete the instrument column - will always be "swe".
- delete kernel set - we dont care about the spice kernel used
- remove the automatic index on kernel_set_key

Same for hit, swapi, codice

src/imap_mag/cli/fetch/ialirt.py
```diff
@@ -64,65 +79,81 @@ def _publish_files(
...
-    downloaded_ialirt: dict[Path, IALiRTPathHandler] = fetch.download_mag_to_csv(
+    downloaded_ialirt: dict[Path, IFilePathHandler] = fetch.download_instrument_data(
+        instrument=instrument,
         start_date=start_date,
         end_date=end_date,
     )
 
     if not downloaded_ialirt:
-        logger.info(f"No I-ALiRT MAG data downloaded from {start_date} to {end_date}.")
+        logger.info(f"No I-ALiRT {instrument.upper()} data downloaded.")
```
alastairtree: Any reason to strip the dates from this log message?

src/imap_mag/client/IALiRTApiClient.py
```diff
@@ -66,7 +66,15 @@ def get_all_by_dates(
...
                 )
-                latest_date = max_chunk_date + timedelta(seconds=1)
+
+                next_date = max_chunk_date + timedelta(seconds=1)
+
+                # avoid getting stuck if I-ALiRT keeps returning same data without advancing in time
+                if next_date <= latest_date:
+                    latest_date = end_date_this_chunk
+                else:
+                    latest_date = next_date
```
alastairtree: next_date, latest_date and end_date_this_chunk - anything you can do to make this easier to understand?! And is there a reason this new check is needed/why new instruments mean it gets stuck now?

src/imap_mag/data_pipelines/DownloadIALiRTStage.py
```diff
@@ -0,0 +1,70 @@
...
+        start_date = getattr(item, "start_date", None)
+        end_date = getattr(item, "end_date", None)
+
+        if not item or not start_date or not end_date:
+            raise ValueError(
+                "DownloadIALiRTStage requires a Record with start_date and end_date"
+            )
+
+        start_date = start_date
+        end_date = end_date
```
alastairtree: This just self assignment? Delete lines?

src/imap_mag/download/FetchIALiRT.py
```diff
@@ -37,33 +37,38 @@ def __init__(
...
         self,
+        instrument: str,
         start_date: datetime,
         end_date: datetime,
-    ) -> dict[Path, IALiRTPathHandler]:
-        """Retrieve I-ALiRT MAG science data."""
+    ) -> dict[Path, IFilePathHandler]:
+        """Retrieve I-ALiRT science data for a specific instrument."""
+
+        processing_map = {"mag": lambda df: process_ialirt_mag_data(df)}
```
alastairtree: Does this still need special functions for each instrument? What will this do with the other instruments?

src/imap_mag/io/file/IALiRTPathHandler.py
```diff
@@ -37,7 +38,14 @@ def get_filename(self) -> str:
         super()._check_property_values("file name", ["content_date"])
         assert self.content_date
 
-        return f"{self.mission}_ialirt_{self.content_date.strftime('%Y%m%d')}.{self.extension}"
+        date_str = self.content_date.strftime("%Y%m%d")
+
+        # if instrument is mag, skip it in the name while other instruments include it
+        # e.g. mag data: imap_ialirt_20260201.csv while swe data: imap_ialirt_swe_20260201.csv
+        if self.instrument.lower() == "mag":
```
alastairtree: I know the current files do not have mag in the name but going forwards I think it would make sense to have mag in the name. The problem is all the existing data files with the old names which are in the datastore/postgres and synced to box. 
I suggest you add a new data migration that does
- find and rename of all the files in the datastore from imap_ialirt_[date].csv to imap_ialirt_mag_[date].csv
- find all file records in the database that match the old naming pattern and assign them a deletion_date. (this will get them removed from box automatically)
- next time the datastore indexer runs it will automatically find all the renamed files and readd them to the database and reupload them to box with the new name
Once that is run on deploy you can assume all data in the system uses the new naming convention of mag in the filename and you can remove the `if self.instrument.lower() == "mag":` branch in this path handler. Sound ok?

src/imap_mag/download/FetchIALiRT.py
```diff
@@ -37,33 +37,38 @@ def __init__(
...
-                content_date=content_date
+                content_date=content_date, instrument=instrument
             ),
-            process_fn=lambda df: process_ialirt_mag_data(df),
+            process_fn=process_fn,
             max_hours_per_chunk=4,  # Limit to 4 hours per chunk to avoid 400 error
         )
 
-    def download_mag_hk_to_csv(
+    def download_instrument_hk_data(
```
alastairtree: Can you simplify `download_instrument_hk_data` and `download_instrument_data` to reduce the duplication?

src/imap_mag/util/constants.py
```diff
@@ -27,6 +27,28 @@ class ENV_VAR_NAMES:
         SDC_URL = "IMAP_DATA_ACCESS_URL"
 
     class DATABASE:
-        IALIRT_PROGRESS_ID = "MAG_IALIRT"
-        IALIRT_HK_PROGRESS_ID = "MAG_IALIRT_HK"
+        IALIRT_PROGRESS_ID = "IALIRT_SCIENCE"
+        IALIRT_HK_PROGRESS_ID = "IALIRT_HK"
         IALIRT_VALIDATION_ID = "IALIRT_VALIDATION"
+
+    class INSTRUMENTS:
```
alastairtree:
```suggestion
    class IALIRT_INSTRUMENTS:
```

src/imap_mag/util/constants.py
```diff
@@ -27,6 +27,28 @@ class ENV_VAR_NAMES:
         SDC_URL = "IMAP_DATA_ACCESS_URL"
 
     class DATABASE:
-        IALIRT_PROGRESS_ID = "MAG_IALIRT"
-        IALIRT_HK_PROGRESS_ID = "MAG_IALIRT_HK"
+        IALIRT_PROGRESS_ID = "IALIRT_SCIENCE"
+        IALIRT_HK_PROGRESS_ID = "IALIRT_HK"
```
alastairtree: When you put this live with changed progress_id keys then the system will think it has not ever downloaded ialirt data before and so will start downloading data from the start of the mission. Is that what you want? If you want it to continue from the current download position (within the last hour usually) then the progress id keys for mag and mag_hk pipelines need to stay the same.

src/prefect_server/pollIALiRT.py
```diff
@@ -26,71 +33,99 @@ def generate_flow_run_name() -> str:
...
+    logger.info(f"Building and running pipeline for {instrument.upper()}...")
+    pipeline.build(run_parameters)
+    await pipeline.run()
+
+    result = pipeline.get_results()
+
+    if not result.success:
+        raise RuntimeError(f"I-ALiRT Pipeline failed for {instrument}: {result}")
+
+    logger = try_get_prefect_logger(__name__)
```
alastairtree, line 83: logger is created here but exists already 9 lines above?

src/prefect_server/pollIALiRT.py
```diff
@@ -128,41 +165,85 @@ async def poll_ialirt_flow(
...
-        while (end_date - DatetimeProvider.now()).total_seconds() > timeout:
-            files = do_poll_ialirt(
-                database, auth_code, start_date, end_date, force_download, logger
-            )
-            updated_files.extend(files)
+        iteration = 1
+        while True:
+            current_time = DatetimeProvider.now()
+
+            if (end_date - current_time).total_seconds() < timeout_seconds:
```
alastairtree: I might be missunderstanding this - if so sorry. But it looks like timeout_seconds=5m so this exits after running for 5m? That is not like the current model or what i specced out with one of each pipeline running every 5mins for 55min, and then the flow restarts hourly?

src/prefect_server/pollIALiRT.py
```diff
@@ -26,71 +33,99 @@ def generate_flow_run_name() -> str:
...
+        event: Event | None = emit_event(
+            event=PREFECT_CONSTANTS.EVENT.IALIRT_UPDATED,
+            resource={
+                "prefect.resource.id": f"prefect.flow-run.{flow_run.id}",
+                "prefect.resource.name": flow_run.name,
+                "prefect.resource.role": "flow-run",
+            },
+            payload={"instrument": instrument, "status": "completed"},
+        )
+        if event is None:
```
alastairtree: These 2 event emit blocks are almost identical and could be simplified

src/prefect_server/pollIALiRT.py
```diff
@@ -128,41 +165,85 @@ async def poll_ialirt_flow(
...
+                logger.info(f"Polling for {inst.upper()}")
+                try:
+                    await run_ialirt_polling_pipeline_task(
+                        instrument=inst,
+                        run_parameters=run_parameters,
+                        database=database,
+                        settings=settings,
+                    )
+                except Exception as e:
+                    logger.error(f"Download failed for {inst.upper()}: {e}.")
```
alastairtree: If all the instruments pipelines fail I think the whole flow should just raise the exception and fail rather than repeating them every 5 minutes so suggest you count exceptions and raise if higher than `len(combined_instruments)`

src/prefect_server/pollIALiRT.py
```diff
@@ -128,41 +165,85 @@ async def poll_ialirt_flow(
...
+
+            # Calculate how long the downloads took
+            batch_duration = (DatetimeProvider.now() - current_time).total_seconds()
+
+            # Sleep only for the remainder of the 5 minutes
+            time_to_sleep = max(0, timeout_seconds - batch_duration)
 
             logger.info(
-                f"--- Waiting {timeout} seconds before polling for new data ---"
+                f"Batch #{iteration} complete in {batch_duration:.1f}s. Sleeping for {time_to_sleep:.1f}s"
```
alastairtree: Nice useful comment thanks!

src/prefect_server/pollIALiRT.py
```diff
@@ -128,41 +165,85 @@ async def poll_ialirt_flow(
...
+            iteration += 1
+    else:
+        for inst in combined_instruments:
+            logger.info(f"Polling for {inst.upper()}")
+            await run_ialirt_polling_pipeline_task(
+                instrument=inst,
+                run_parameters=run_parameters,
+                database=database,
+                settings=settings,
+            )
```
alastairtree: Minor nitpick here (from the inherited code too sorry!) but i think i would have the loop above run one in all cases and exit early if `wait_for_new_data_to_arrive=True` rather than having an if branch and 2 main for loops. Reduces duplication and less chance of a later refactor changing one version and not the other.

src/prefect_server/workflow.py
```diff
@@ -300,13 +293,6 @@ async def adeploy_flows(local_debug: bool = False):
                     "prefect.resource.name": PREFECT_CONSTANTS.FLOW_NAMES.POLL_IALIRT
                 },
             ),
-            DeploymentEventTrigger(
-                name="Trigger upload after I-ALiRT HK poll",
-                expect={PREFECT_CONSTANTS.EVENT.FLOW_RUN_COMPLETED},
-                match_related={
-                    "prefect.resource.name": PREFECT_CONSTANTS.FLOW_NAMES.POLL_IALIRT_HK
-                },
-            ),
```
alastairtree, line 344: Since there is now only event from IALIRT that triggers SHAREPOINT_UPLOAD flow, I think you only need to emit one event from the ialirt flow and the 2nd event can be removed

src/imap_mag/download/FetchIALiRT.py
```diff
@@ -37,33 +37,38 @@ def __init__(
...
         self,
+        instrument: str,
         start_date: datetime,
         end_date: datetime,
-    ) -> dict[Path, IALiRTHKPathHandler]:
+    ) -> dict[Path, IFilePathHandler]:
```
alastairtree: Could we have a single path handler for all I-ALiRT files now rather than falling back to the low level IFilePathHandler? And so just get rid of  `IALiRTHKPathHandler` and use `IALiRTPathHandler` for both? Might be a nice improvement?

tests/test_pollIALiRT.py
```diff
@@ -349,76 +349,3 @@ async def test_poll_ialirt_send_quicklook_at_6am_uk_time(
...
     # Verify.
     mock_teams_webhook_block.notify.assert_called_once()
-
-
-@pytest.mark.skipif(
-    os.getenv("GITHUB_ACTIONS") and os.getenv("RUNNER_OS") == "Windows",
-    reason="Wiremock test containers will not work on Windows Github Runner",
-)
-@pytest.mark.asyncio
-async def test_poll_ialirt_hk_autoflow_first_ever_run(
```
alastairtree, line 355: These 2 tests have been deleted but they look like they are using wiremock and probably do exercise the hk branch of code in the pipeline so might still be relevant?

alastairtree (commented):
Thank very @miruuna this is great work. And sorry this took me so long to get back to you. I have left you a few improvement/refactor suggestions in inline comments but you have solved all the big issues here nicely. I have tests the CLI commands with our prod key for all the instruments and the mag hk files and all work nicely thanks.

- Please could you add the new CLI options to the readme as well.

Thank you, epic stuff!

alastairtree:
@copilot this branch is out of date and i want you to make a PR to update it. Create a new branch based on this branch and then merge main into it. Fix all the issues and conflicts and update any code to use the new date provider logic recently introduced in main. Ensure the tests and QA checks pass. Create a PR of your work into this branch feat/create-a-pipeline-to-azure-for-i-alirt-hit-swe-swapi-and-codice-data and assign @miruuna to review it

src/imap_mag/download/FetchIALiRT.py
miruuna: This should not do anything to the data from other instruments as there is currently no further processing implemented for those, but left it in there in case we'll have to add new functions for processing.

 I've added a check in `__download_to_csv()`  to check if a `process_fn` is passed as an arg.

src/prefect_server/pollIALiRT.py
miruuna: I think the naming was confusing. `timeout_seconds` was actually the sleep between the batches, not the flow. I have switched to launching the tasks concurrently, as it seems that using `return_exceptions=True` prevents all tasks from failing if one of them is failing (my fear was that they'd all fail, hence why I went for sequential tasks). I think I misunderstood it a bit  but I changed it so that the flow stops at 55 min instead of 59 and launches tasks every 5 minutes.

src/prefect_server/pollIALiRT.py
miruuna: That is a good idea. Changed it so now it raises a Prefect `Failed`  when all instruments in a batch fail.

src/prefect_server/pollIALiRT.py
miruuna: Did you mean when `wait_for_new_data_to_arrive=False`? That makes sense! I refactored it to reduce it to one loop only.

src/prefect_server/workflow.py
miruuna: I missed that, thanks for spotting. I removed one of the triggers.

src/imap_mag/io/file/IALiRTPathHandler.py
miruuna: Added a new migration file using the previous migration as an example. Tested it locally and it works as intended.

miruuna: However, do the existing hk files follow the same naming pattern?

alastairtree: Yes they do - today’s file is like this imap_ialirt_hk_20260626.csv and i guess to be consitent it also needs mag. good spot.

src/imap_mag/download/FetchIALiRT.py
alastairtree: Ah ok i had forgotten that it to add extra processing - thanks for the reminder

miruuna:
hi @alastairtree I think I have now addressed all of the PR comments. I have a couple of things I am still unclear about:
 -  `IALiRTHKPathHandler` does not exist anymore and is now replaced by `IALiRTPathHandler(is_hk=True)`. Therefore, mag_hk files will now be stored in `../ialirt/..` rather than where they were previously stored in `../ialirt_hk/..`. Should the hk files currently in the DataStore be move to `ialirt/`?
 - I am not convinced my implementation of the `DateTimeProvider` in `pollIALiRT.py` is correct as using `fixed_now=NOW` doesn't have any effect and the actual time is still being used.

alastairtree:
Yes ialirt_hk files need to move to be be consistent with the new files being generated and placed into new folders going forwards. Or the handler should put the new files in the old ialirt_hk folder. But let's keep it consistent. How are all the files for alirt arranged now, can you give some sample paths for the different files?

Not sure what you mean about date time provider but I will take a look. Sorry that got massively refactored to make it an instance class not static because otherwise tests on different threads were retiming each others clocks.

miruuna:
> Yes ialirt_hk files need to move to be be consistent with the new files being generated and placed into new folders going forwards. Or the handler should put the new files in the old ialirt_hk folder. But let's keep it consistent. How are all the files for alirt arranged now, can you give some sample paths for the different files?
> 
> Not sure what you mean about date time provider but I will take a look. Sorry that door massively refactored to make it an instance class not static because otherwise tests on different threads were retiming each others clocks.

Ah sorry I should have been more clear. The paths  right now would be:
-  for science instruments: `ialirt/2025/10/imap_ialirt_mag_20251014.csv`, `ialirt/2025/10/imap_ialirt_swe_20251014.csv`, etc
- for mag hk: `ialirt/2025/10/imap_ialirt_mag_hk_20251014.csv`

As for the `datetime_provider`, I am unsure if the flow in  [src/prefect_server/pollIALiRT.py ](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/7f60e8591930fdfc4d7291f1c8c6f3a2ae4d2ed4/src/prefect_server/pollIALiRT.py) uses it properly. It came up in the integration tests in [test_pollIALiRT.py](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/7f60e8591930fdfc4d7291f1c8c6f3a2ae4d2ed4/tests/integration/test_pollIALiRT.py) when I couldn't use the fixed `NOW` as the pipeline still uses the actual now timestamp to update the db.

alastairtree:
I have fixed your use of the datetime provider - you had it correctly setup for the tests but if you actually run the flow then it was always null so I have assigned one in that case. It is annoying to have to pass it as a flow parameter and have it show up in the UI but I could not find a better way to do dependency injection in flows (classes do not seem to work). Need it so it could be stubbed/fixed-time in tests (and that can run in parallel) while also just working like datetime.utcnow in prod. 
Open to suggestions on a better way ;-).

src/imap_mag/data_pipelines/IALiRTInstrumentPipeline.py
```diff
@@ -34,9 +34,9 @@ def __init__(
         self.base_instrument = instrument.replace("_hk", "")
 
         if self.is_hk:
-            progress_id = f"{self.base_instrument.upper()}_{CONSTANTS.DATABASE.IALIRT_HK_PROGRESS_ID}"
+            progress_id = f"{CONSTANTS.DATABASE.IALIRT_HK_PROGRESS_ID}"
         else:
-            progress_id = f"{self.base_instrument.upper()}_{CONSTANTS.DATABASE.IALIRT_PROGRESS_ID}"
+            progress_id = f"{CONSTANTS.DATABASE.IALIRT_PROGRESS_ID}"
```
alastairtree: Does this mean their progress for all instruements is tracked under one key `MAG_IALIRT` - what happens if one if one instrument has no data because it is off? How to we maintain up to date data with one while the other would have a much older progress date?

src/imap_mag/io/file/IALiRTPathHandler.py
alastairtree, line 1: Nice refactor to merge those two - lots of duplicated code removed thanks

src/imap_mag/util/constants.py
```diff
@@ -27,11 +27,11 @@ class ENV_VAR_NAMES:
         SDC_URL = "IMAP_DATA_ACCESS_URL"
 
     class DATABASE:
-        IALIRT_PROGRESS_ID = "IALIRT_SCIENCE"
-        IALIRT_HK_PROGRESS_ID = "IALIRT_HK"
+        IALIRT_PROGRESS_ID = "MAG_IALIRT"
+        IALIRT_HK_PROGRESS_ID = "MAG_IALIRT_HK"
```
alastairtree, line 33: Changing these means the existing progress records in the prod DB will be lost and it will redownload everything from the start of the mission - is that the plan?

src/prefect_server/pollIALiRT.py
```diff
@@ -194,61 +191,72 @@ async def poll_ialirt_flow(
...
+        time_left_in_flow = (end_date - datetime_provider.now()).total_seconds()
+        if time_to_sleep > time_left_in_flow:
+            time_to_sleep = max(0, time_left_in_flow)
+
+        logger.info(
+            f"Batch #{iteration} complete in {batch_duration:.1f}s. Sleeping for {time_to_sleep:.1f}s"
+        )
+
+        await asyncio.sleep(time_to_sleep)
+        iteration += 1
```
alastairtree, line 248: Loop looks much cleaner thanks!

alastairtree (commented):
Looking good thank you! Spotted one issue where I think you are merging all instruments workflow progress records together maybe?

Not clear where/if the HK files are going to be moved?

Have you tested it in the test env? If not please do that. And maybe tomorrow we could have a call where you do a demo of it working?

src/imap_mag/util/constants.py
miruuna: You are right, we don't want to change those. This change is to revert back to the original IDs, which are `MAG_IALIRT` and `MAG_IALIRT_HK`

src/imap_mag/data_pipelines/IALiRTInstrumentPipeline.py
miruuna: You are right, we'd want to have an id for each instrument. I've gone back to the previous version.

miruuna:
Thanks for adding the fixes @alastairtree. I've only tried it on a local dev environment (with a local docker-based DB) but I currently don't have the IALIRT API secrets so can't run the flow to test.

As for the HK files, I decided to revert to storing them in `ialirt_hk` for clarity - now the paths look like  `ialirt_hk/2025/10/imap_ialirt_mag_hk_20251014.csv`

src/prefect_server/pollIALiRT.py
```diff
@@ -1,192 +1,120 @@
...
-        else DatetimeProvider().end_of_hour()
+    pipeline = IALiRTPipeline(
+        instrument=instrument, database=database, settings=settings
     )
 
-    return (
-        f"Poll-IALiRT-HK-from-{start_date}-to-{end_date.strftime('%d-%m-%YT%H:%M:%S')}"
+    logger.info(f"Building and running pipeline for {instrument.upper()}...")
+    pipeline.build(run_parameters)
+    await pipeline.run()
```
alastairtree, line 76: I saw the concurrency error in prefect on mag-agent. We see these a fair bit - it is a stupid bug in the version of prefect where all the CPU gets locked in the flow and so the background process that renews a lease with the server gets stuck so it breaks it's lease and marks an error. Have a go with changing this to something like this using anyio to_thread:
```python
from anyio import to_thread
....
await to_thread.run_sync(lambda: pipeline.run())
```
Then try that out with a big download/long flow and see if that helps fix it

miruuna: I tried changing that but the tasks get stuck with `Task run failed with exception: RuntimeError('Cannot get results from pipeline that has not completed.') `. It would probably need some more changes to make it work.

I reverted to `await` and reduced the `max_hours_per_chunk` to `1.5` hours just for the HK data and it seems to work fine.

alastairtree:
I have been testing this in the MAG-AGENT test env running with the default args (like the hourly scheduled jobs will run) and spotted some bugs:
1. The database upload has not been tested as there are no tables for ialirt data from other instruments in the database. There was no new iairt data in the database. i ran a job to import the new ialirt files created with the updated config but it gave an error - look at http://mag-agent.ph.ic.ac.uk/runs/flow-run/2cb78082-af18-4f11-8d43-a36d7839054e
2. The jobs run every 5 minutes with tasks which is nice, but i think they are constantly redownloading the same data and not only fetching the new data. For example after running many times in one hour in [this job run](http://mag-agent.ph.ic.ac.uk/runs/flow-run/1bb1789d-477f-4415-bff6-873116374fc7?tab=Logs) it was still emitting this into the logs - why is it starting at midnight when it has already run earlier on in the hour and downloaded some data for today? each one should be updating the workflow progress with the latests data found so it only downloads new data I think?
`Downloading I-ALiRT mag_hk data from 2026-07-09 00:00:00 to 2026-07-09 23:59:59.999999.`
3. It seems to make more queries than it needs - it is downloading in time windows that are in the future e.g. running at 2pm on 2026-07-09 it was making these requests which are pointless as they are in the future and that data cannot ever exist yet.
```
Algorithm query: GET https://ialirt.imap-mission.com/api-key/space-weather?instrument=codice_hi&time_utc_start=2026-07-09T16%3A04%3A03&time_utc_end=2026-07-09T20%3A04%3A03
Algorithm query: GET https://ialirt.imap-mission.com/api-key/space-weather?instrument=codice_hi&time_utc_start=2026-07-09T20%3A04%3A03&time_utc_end=2026-07-09T23%3A59%3A59
```

Compare the polling pattern in your code with these logs from our existing HK process - it only gets data from the last run up until "now"

image

Think this needs a bit more work and some more testing.

miruuna:
I have added the fixes and it should now work closer to what was intended:
Scenario 1: 
-  if no start and end dates provided by the user, the **flow** uses the `progress_timestamp` from the db as `start_date` and the end of the hour as `end_date`
- the flow launches  a batch of tasks (a task for each instrument),every 5 minutes, so the data is fetched from the API in 5 minutes increments - no more spamming the api for future dates

Scenario 2:
- user provides start and end dates in the past to fill up historical gaps
- the flow launches a single batch of tasks that will download the data in increments  equal to the `min(max_hours_per_chunk, time_left_until_end_date)

The uploading to the database also seems to work fine, but it hits some errors when saving some of the `codice` data. This was to be expected since we don't currently have any processing for any instruments apart from `mag` and `mag_hk`

alastairtree:
Sounds good - have looked at the runs you did in on mag-agent and they look much better thanks!
I was expecting all the other instrument data to be processed and appear in the database, anything you can do to get it working?
Have merged in the latest main to your branch to keep it up to date.

miruuna:
> Sounds good - have looked at the runs you did in on mag-agent and they look much better thanks! I was expecting all the other instrument data to be processed and appear in the database, anything you can do to get it working? Have merged in the latest main to your branch to keep it up to date.

As discussed, I decided to remove the codice instruments from the pipeline for now as we won't need them for the website and their data require further processing. This can be tackled in a different issue.

I have tested in the test env and the `pollIALiRT` flow is working as intended - http://mag-agent.ph.ic.ac.uk/runs/flow-run/06f44fab-bce6-4b2d-88a0-097b10ebb64e?tab=Task+Runs 


The instrument data is uploaded in the respective tables: `ialirt_mag`, `ialirt_mag_hk`, `ialirt_hit`, `ialirt_swapi` and `ialirt_swe`.
Screenshot from 2026-07-21 13-41-50

Screenshot from 2026-07-21 13-55-27

Checked that the file renaming migration script worked - the old mag files were marked for deletion (migration was done at 9:29)
Screenshot from 2026-07-21 14-02-50

alastairtree:
Looks good thanks. I will merge and deploy. 
Go live plan:
- merge and release new version 
- Diff database schema in test and prod to create a delta script. Run the delta script against the azure database to migrate it before crump updates prod database
- grant read on the new tables to the website database user
- pause all jobs
- deploy change to prod in a new version
- remove the ialirt hk deployment from prod as this has been combined
- all existing IALIRT files will be renamed by migration so check this works on deploy. They will be set as deleted in database
- Run the indexer after deploy to reindex the csv files
- ensure ialirt updated files are deleted/reuploaded to box
- run the ialirt job for the last 7 days
- add the new tables to the replication so results appear in azure
- update the website so data appears

alastairtree:
SQL migration script we will need in azure:
```sql
CREATE TABLE IF NOT EXISTS public.ialirt_hit
(
    id timestamp without time zone NOT NULL,
    hit_e_a_side_med_en integer NOT NULL,
    hit_e_b_side_med_en integer NOT NULL,
    hit_epoch bigint NOT NULL,
    hit_h_omni_low_en integer NOT NULL,
    hit_h_omni_med_en integer NOT NULL,
    hit_he_omni_high_en integer NOT NULL,
    hit_he_omni_low_en integer NOT NULL,
    ttj2000ns bigint NOT NULL,
    file_date date,
    CONSTRAINT ialirt_hit_pkey PRIMARY KEY (id)
)

TABLESPACE pg_default;

GRANT SELECT ON TABLE public.ialirt_hit TO azure_sdvdb_repl_user;

ALTER TABLE IF EXISTS public.ialirt_hit
    OWNER to "imap-db";

GRANT ALL ON TABLE public.ialirt_hit TO "imap-db";



CREATE TABLE IF NOT EXISTS public.ialirt_swapi
(
    id timestamp without time zone NOT NULL,
    swapi_epoch bigint NOT NULL,
    swapi_pseudo_proton_density double precision NOT NULL,
    swapi_pseudo_proton_speed double precision NOT NULL,
    swapi_pseudo_proton_temperature double precision NOT NULL,
    ttj2000ns bigint NOT NULL,
    file_date date,
    CONSTRAINT ialirt_swapi_pkey PRIMARY KEY (id)
)


TABLESPACE pg_default;

GRANT SELECT ON TABLE public.ialirt_swapi TO azure_sdvdb_repl_user;

ALTER TABLE IF EXISTS public.ialirt_swapi
    OWNER to "imap-db";

GRANT ALL ON TABLE public.ialirt_swapi TO "imap-db";


CREATE TABLE IF NOT EXISTS public.ialirt_swe
(
    id timestamp without time zone NOT NULL,
    swe_counterstreaming_electrons integer NOT NULL,
    swe_epoch bigint NOT NULL,
    swe_normalized_counts character varying(81) COLLATE pg_catalog."default" NOT NULL,
    ttj2000ns bigint NOT NULL,
    file_date date,
    CONSTRAINT ialirt_swe_pkey PRIMARY KEY (id)
)

TABLESPACE pg_default;

GRANT SELECT ON TABLE public.ialirt_swe TO azure_sdvdb_repl_user;

ALTER TABLE IF EXISTS public.ialirt_swe
    OWNER to "imap-db";

GRANT ALL ON TABLE public.ialirt_swe TO "imap-db";


GRANT SELECT ON TABLE public.ialirt_hit TO sdvwebsiteuser;
GRANT SELECT ON TABLE public.ialirt_swapi TO sdvwebsiteuser;
GRANT SELECT ON TABLE public.ialirt_swe TO sdvwebsiteuser;
```

alastairtree:
This is now in prod

PR #303, opened 2026-06-17
Claude/improve code coverage

alastairtree:
- Fix datetime provider so it is an instance and not a globally shared static. Means tests do not interfere with each other across threads
- Adds lots of tests
- Splits tests into unit tests (fast - pure python only) and integration (slower - needs prefrect server or database or wiremock)

Test run in 2.5min locally and 4m in actions which is faster than the current main despite so many extra tests

Coverage > 90% now


alastairtree:
@mhairifin this might be a tough one to review - maybe just review the code and not the tests? I think the tests are basically just bonus coverage as we discussed this week.

src/imap_mag/util/DatetimeProvider.py
```diff
@@ -5,43 +5,44 @@
...
-    def end_of_today(cls) -> datetime:
-        return cls.today().replace(hour=23, minute=59, second=59, microsecond=999999)
+    def end_of_today(self) -> datetime:
+        return self.today().replace(hour=23, minute=59, second=59, microsecond=999999)
 
-    @classmethod
-    def beginning_of_imap(cls, date_type: type[T] = datetime) -> T:
-        return cls.today(date_type).replace(year=2025, month=9, day=24)
+    def beginning_of_imap(self, date_type: type[T] = datetime) -> T:
+        return self.today(date_type).replace(year=2025, month=9, day=24)
```
mhairifin: I think this date should maybe be stored in a constant somewhere rather than buried in a class method - I also think it should probably just return the start of mission date directly rather than doing today.replace - I don't know what the purpose of that is

src/prefect_server/pollIALiRT.py
```diff
@@ -421,3 +151,315 @@ def _do_poll(
...
+                force_latest_update=True,
+            )
+
+            # If this is the 6 AM (UK time) polling job, send the latest figure to Teams
+            uk_end_time = end_date.replace(tzinfo=UTC).astimezone(
+                pytz.timezone("Europe/London")
+            )
+
+            if wait_for_new_data_to_arrive and (
+                uk_end_time.hour == 6 or force_notification
```
mhairifin: I know this is a refactor of old code that also did this - but I do not think the 6 should be hard coded here. Should at the least be a constant somewhere, and should probably be an input parameter to the flow that defaults to 6 or similar?

src/prefect_server/pollScience.py
```diff
@@ -283,3 +87,213 @@ def get_latest_ingestion_date(downloaded_science):
     )
 
     return latest_ingestion_date
+
+
+class PollScienceFlow:
```
mhairifin: A lot of these poll flows have very similar class signatures - perhaps not for now but I wander if it would be good for them to eg inherit from an interface, particularly as I think that then informs an easy self documenting template for building new flows

tests/util/miscellaneous.py
```diff
@@ -11,14 +11,16 @@
...
 from imap_mag.util import DatetimeProvider, Environment
 
-NOW = datetime(2025, 6, 3, 12, 37, 9)  # should not be first Monday of month
-TODAY = NOW.replace(hour=0, minute=0, second=0, microsecond=0)  # 2025-06-03
-TOMORROW = TODAY + timedelta(days=1)  # 2025-06-04
-YESTERDAY = TODAY - timedelta(days=1)  # 2025-06-02
-START_OF_HOUR = NOW.replace(hour=0, minute=0, second=0, microsecond=0)
+NOW = datetime(
+    2025, 10, 14, 12, 37, 9
+)  # should not be first Monday of month; after IMAP launch
```
mhairifin: Why not first Monday of the month?

mhairifin (approved):
Overall looks good - a few comments on the code

src/imap_mag/util/DatetimeProvider.py
alastairtree: Fixed

src/prefect_server/pollScience.py
alastairtree: Had to revert all the class based stuff - prefect starting flows via docker cannot cope with them annoyingly. Pure functions only

Issue #305, opened 2026-06-18
Ingest SOLAR-1, ACE and DSCOVR data


dalonsoa:
The magnetic and solar wind data for SOLAR-1, ACE and DSCOVR is provided in several JSON files updated every minute for the last 24h. These data should be ingested in the database and this one being used by the web app, rather than hooking into the different servers and processing the files at the web app side of things. 

Specifically, addressing this will require:

## Download and processing

Conceptually, these would be similar to [this one](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/download/FetchIALiRT.py), fetching the data from the relevant place and then processing it into a CSV file so we store only the relevant information. I guess we need a [CLI to fetch the data](https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/cli/fetch/fetch.py) as well, but do we need to setup a [stage](https://github.com/ImperialCollegeLondon/imap-pipeline-core/tree/main/src/imap_mag/data_pipelines), as well?

- Create a pipeline to download https://services.swpc.noaa.gov/json/rtsw/rtsw_mag_1m.json , which contains MAG data for SOLAR-1 and ACE for the last 24h.
- Create a pipeline to download https://services.swpc.noaa.gov/products/solar-wind/mag-2-hour.json which contains MAG data for DSCOVR for the last 2h. There is also a 5 min file, but we might very well find it empty if we try to download it at the wrong time. Alternatively, there's a 1 day file.
- Create a pipeline to download https://services.swpc.noaa.gov/json/rtsw/rtsw_plasma_1m.json , which contains solar wind data for SOLAR-1 and ACE for the last 24h.
- Create a pipeline to download https://services.swpc.noaa.gov/products/solar-wind/plasma-2-hour.json which contains solar wind data for DSCOVR for the last 2h. There is also a 5 min file, but we might very well find it empty if we try to download it at the wrong time. Alternatively, there's a 1 day file.

We might want to do some processing, like calculating phi and theta for DSCOVR, which is not available.

Probably, we want to have just 1 file in the end with 4 columns: datetime, spacecraft, variable, value and ingest that in a single model with the same fields.

## Ingesting into the DB

To ingest the data into the database, we need, for the data in the four files generated above:
[...]

PR #308, opened 2026-06-23
Add client to pull data for other L1 spacecrafts

### LINKED ISSUE #339: Create client to pull data from NOAA's RTSW endpoints
Should require spacecraft and instrument as input, and only work for `SOLAR1` and `ACE`, and for `mag` and `wind`, respectively.

Fixed by #308

dalonsoa:
Adds the client, and a few configuration options that will be used later in the fetcher, to download the data for other L1 spacecrafts. For now, SOLAR-1, ACE and DSCOVR. 

Appropriate tests to check the functionality will be added when implementing the fecther.

This is the first step of #305. Next it will be actually fetching and processing the data, saving it to CSV files, and, finally, ingesting them into the DB.


dalonsoa:
@alastairtree I thought it best to break the issue into smaller chunks (aka PR): client, fetcher, CLI and hooking everything together. Let me know if this works for you or if you prefer an all-or-nothing kind of PR.

imap-mag-config.yaml
```diff
@@ -4,12 +4,12 @@
 # and by dictionaries passed into the constructor of the AppSettings class
 
 # General application conifguration:
-work_folder: '.work'
-data_store:  'tests/datastore'
-packet_definition: 'packet_def/'
+work_folder: ".work"
```
alastairtree: I dont actually care but have a hunch that YAML is "supposed" to be single quotes by default. Presume you have a YAML IDE plugin that has "fixed" these? anyway a nitpick  - and you may have not meant to do it - doubt it matters!

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+from datetime import datetime
+from time import time
+from typing import Any
+
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class OtherL1ApiClient:
```
alastairtree: Maybe name this NoaaApiClient rather than "other"?

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+    """Interacts with SOLAR-1, ACE and DSCOVR APIs.
+
+    Which specific JSON file to use depends on the spacecraft as well as the
+    time range to retrieve, using the file that covers as much as possible
+    within the requested time range.
+    """
+
+    def __init__(self, solar1_ace_url: str, dscovr_url: str):
```
alastairtree: Would be tempted to pass in the whole settings object here rather than the URLs as a nod to the open/closed principal.

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+
+        Args:
+            base_url: The base URL of the API.
+            file_name: The name of the JSON file to download.
+
+        Returns:
+            A list of dictionaries containing the JSON data.
+        """
+
+        url = f"{base_url}/{file_name}"
```
alastairtree:
````
url = f"{base_url.rstrip('/')}/{file_name.lstrip('/')}"
```

Suggest this to catch config oopsies with double slashes

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+            response = requests.get(url)
+            response.raise_for_status()  # Raise an error for bad responses
+            return response.json()  # Assuming the response is a JSON array
+        except requests.RequestException as e:
+            logger.error(f"Failed to download {url}: {e}")
+            raise
+
+    def _get_solar1_ace_data(self) -> list[dict[str, Any]]:
+        """Download SOLAR-1 and ACE data.
+
```
alastairtree: Worth adding a comment that this is ther RTSW / real time space weather api?

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+    """Interacts with SOLAR-1, ACE and DSCOVR APIs.
+
+    Which specific JSON file to use depends on the spacecraft as well as the
+    time range to retrieve, using the file that covers as much as possible
+    within the requested time range.
+    """
+
+    def __init__(self, solar1_ace_url: str, dscovr_url: str):
+        self._solar1_ace_url: str = solar1_ace_url
+        self._dscovr_url: str = dscovr_url
```
alastairtree: Add some validation here for empty config values?

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+        try:
+            mag_data: list[dict[str, Any]] = self._download_json_file(
+                self._solar1_ace_url, mag_file
+            )
+            plasma_data: list[dict[str, Any]] = self._download_json_file(
+                self._solar1_ace_url, plasma_file
+            )
+        except Exception as e:
+            logger.error(f"Error downloading SOLAR-1 and ACE data: {e}")
+            return []
```
alastairtree: Looks like this is quietly swallowing API exceptions. I can see the value in downloading and processing both semi-independently - if one API is down it is still useful to get the other (if available) but I think the calling overall prefect flow should fail  and trigger our failures/notifications if one of the the APIs are down (whilst still processing the other API data if thazt ios working ok). Or split into 2 flows and the error handling is easier - just throw? What do you think it should do iof the APIs are erroring?

src/imap_mag/client/NOAAApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+            A list of dictionaries containing the DSCOVR data.
+        """
+        # Determine which file to download based on the time range
+        time_ranges = {
+            "2-hour": 2,
+            "6-hour": 6,
+            "1-day": 24,
+            "3-day": 72,
+            "7-day": 168,
+        }
```
alastairtree, line 135: clever solution!

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+        # Download the data from the DSCOVR API
+        try:
+            mag_data: list[dict[str, Any]] = self._download_json_file(
+                self._dscovr_url, _mag_file
+            )
+            plasma_data: list[dict[str, Any]] = self._download_json_file(
+                self._dscovr_url, _plasma_file
+            )
+        except Exception as e:
+            logger.error(f"Error downloading DSCOVR data: {e}")
```
alastairtree: Same as above on error handling strategy here

src/imap_mag/config/FetchConfig.py
```diff
@@ -30,3 +32,13 @@ class FetchSpiceConfig(CommandConfig):
 class FetchWebTCADLaTiSConfig(CommandConfig):
     api: WebTCADLaTiSApiSource
     publish_to_data_store: bool = True
+
+
+class FetchSOLAR1andACEConfig(CommandConfig):
+    api: SOLAR1andACEApiSource
+    publish_to_data_store: bool = False
```
alastairtree: Suspect you will want this to default to true - means do the CSV files get copied into the main storage area for the pipelines.

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+            base_url: The base URL of the API.
+            file_name: The name of the JSON file to download.
+
+        Returns:
+            A list of dictionaries containing the JSON data.
+        """
+
+        url = f"{base_url}/{file_name}"
+        try:
+            response = requests.get(url)
```
alastairtree: Suggest you add a timeout=30 (or similar) so cannot hang forever. We should probably add that to the other API requests now i think of it! Could also either handle requests.exceptions.Timeout or just add a rerty/delay argument to the flow decorater when it is done.

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+            response.raise_for_status()  # Raise an error for bad responses
+            return response.json()  # Assuming the response is a JSON array
+        except requests.RequestException as e:
```
alastairtree: The caller also has this wrapped in try catch meaning I think this logs exceptions twice?

src/imap_mag/client/OtherL1ApiClient.py
```diff
@@ -0,0 +1,165 @@
...
+    time range to retrieve, using the file that covers as much as possible
+    within the requested time range.
+    """
+
+    def __init__(self, solar1_ace_url: str, dscovr_url: str):
+        self._solar1_ace_url: str = solar1_ace_url
+        self._dscovr_url: str = dscovr_url
+
+    @staticmethod
+    def _download_json_file(base_url: str, file_name: str) -> list[dict[str, Any]]:
```
alastairtree: Copilot spotted this cleverly:
> _download_json_file is annotated -> list[dict[str, Any]], but the DSCOVR endpoint returns list[list] (array of arrays), not dicts. The annotation is only correct for one of two code paths. Fix: change return type to list[Any] to accurately reflect both formats.

alastairtree (commented):
Looking great so far - thank you! Have added a few inline comments, mostly small nitpicky things around error handling but happy path is looking good. Think the merge is also no longer clean - sorry that is my fault.

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: Happy to do that, but that's not how the IALiRT client (or the other clients) seem to be designed.

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: And what if we change the source and pull data from somewhere else? This will always pull L1 data, but not necessarily from NOAA.

This reminds me that it is unclear to me what something should be a configurable input, included in the config file, and what something hardcoded. Sometimes the line is quite blured. For example, we are making the URL for this APIs configurable, but we are hardcoding - or will be - the structure of the response (format of the data, column names, etc. so we can process them). 

Do you have any suggestion on this?

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: If we want to do it truly independent, we would need 4 separate pipelines, one per endpoint. I feel that is quite overkilling, so I rather let the whole thing fail and Prefect to handle the errors and retries. 

To confirm, to do that, I should just log the error and raise again instead of returning an empty list, right?

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: Same response :)

src/imap_mag/config/FetchConfig.py
dalonsoa: I was not sure how that part worked. Thanks for the pointer.

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: True. I leave only the caller, as I think the log message is going to be more informative.

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: Clever Copilot...

imap-mag-config.yaml
dalonsoa: For what I can tell, even before this PR, there are fields in that file that are single quotes and other that are double quotes. In general, it seems that using one or the other depends on what you are quoting. See https://dev.to/darkmavis1980/single-quotes-vs-double-quotes-in-yaml-how-they-differ-from-javascript-and-typescript-20hj

src/imap_mag/client/OtherL1ApiClient.py
alastairtree: URL needs to be configurable because the tests will substitute it for a wiremock stub.

The client will coded for that exact API, cannot see any way round it, so it is a NOAA client. If we need another source we will rewrite it or add another client.

src/imap_mag/client/OtherL1ApiClient.py
alastairtree: Fair point!

dalonsoa:
@alastairtree I've implemented all of your suggestions - except the change name to NOAA, see comment - and made some other changes that I felt were needed, like remove the end date (these APIs always provide data until now). Please, have a look again.

src/imap_mag/client/OtherL1ApiClient.py
dalonsoa: Just seen this. OK, updating.

src/imap_mag/client/OtherL1ApiClient.py
alastairtree: Raised exceptions are logged automatically if they are uncaught so I think just let it raise. 

I think you need separate pipelines anyway, the data may be updated at different cadences -one instrument could be turned off for a week day- so you need to keep track with the workflow progress system of the most recent data you already have for each one so you can only grab the new stuff. Each one needs to keep updating even if the others don't. 

Hopefully using the pipeline infra or will be a lot of reusable code!

dalonsoa: Just to confirm: 4 pipelines mean 4 clients, 4 fetchers, 4 CLI, etc. Right?

dalonsoa: My thought was to dump everything in a single table - the three missions -, with a field indicating the mission. I guess that is not the way to go if we have separate pipelines.

alastairtree: Hmm interesting. I reckon tables in the database that look very similar to the source data and and API structure do as little processing as possible. And let crump do all the dB schema stuff for you.

alastairtree: And just store the CSV files alongside all the other imap data files

dalonsoa: So... what do you prefer? 1 pipeline and 1 table (my original idea) or 4 pipelines and 6 tables (one per mission for plasma and mag)?

alastairtree: If I was doing it I would probably do it based on the source data - one file and db table per API you are fetching from. And then you basically build a general purpose pipeline JSON -> CSV file -> DB (via crump) and reuse it for each one. But I am very happy for you to decide differently. I think the requirement that each dataset still updates even if one of the others is offline/decommissioned is important though. The spacecraft and instruments will definitely change over time as they get turned off, damaged or decommissioned and so the remaining ones still need to work and each stay up to date even when one API is down or returning less data.

alastairtree (approved):

alastairtree:
Looks good. Have commented about the number of pipelines inline. Happy for you to merge when ever you want to as nothing is actually being used and you can edit them as you like in the next PR. if needed Be good to see a wiremock stub for those APIs and some test coverage in the next one as well. Thank you!

dalonsoa:
OK, after some thought, I'm going to re-do this, creating separate clients for the 4 endpoints we are connecting to. As it is now, it just does not allow for the granularity and independence we are looking for. I'll also add the stubs along the way, so this part of the code can be tested properly. No input needed, for now.

dalonsoa:
@alastairtree I've updated the Client, splitting the one we had to handle the different endpoints separately. I've also added tests. Have a look at it and, if you are happy with it, this is ready to go.

alastairtree (approved):

alastairtree:
Lovely - thank you! Look forward to the next stage.

PR #325, opened 2026-07-07
feature: Scripted L2 calibration (MATLAB pipeline)

alastairtree:
## Summary

Adds a new `SCRIPTED_L2_CALIBRATION` calibration method that runs the external
MATLAB `calibration.scripts.calibrate_l2_offsets` pipeline and feeds its layer
output into the existing calibration flow, alongside the existing methods.

## What's included

- **`ScriptedL2CalibrationConfig`** — a `CalibrationConfig` subclass with
  `calibration_matrix_version` and a non-empty `input_json_file`.
- **`ScriptedL2CalibrationJob`** — generates the MATLAB user/env file-path config
  in the work folder (`sharepoint_flight_data` + `spice_metakernal_root` → datastore
  root; the three output folders → work folder), resolves or generates the SPICE
  metakernel, builds and runs `calibrate_l2_offsets` from the MATLAB repo root, and
  collects the produced `imap_mag_manual-{mode}-layer_{date}_v{VVV}` JSON + CSV.
- **`MatlabWrapper.call_matlab`** gains `cwd`, `unset_display` and
  `include_project_paths` so the self-contained external MATLAB project can be
  invoked from its own root with `DISPLAY` unset.
- **`calibrate_flow` / `calibrate`** gain `metakernel` and `matlab_repo` arguments.
  The flow resolves `matlab_repo` (a `LocalFileSystem` block, `GitHubRepository`
  block, a block name, or `None`) to a local path — cloning a GitHub repo into a
  work-folder subfolder named after the repo when needed — then passes the resolved
  path down. `metakernel` is treated like `spice_metakernel` in `apply_flow`.
- Other calibration methods are unchanged.

## Verification

Ran end-to-end (`calibrate_flow` → block resolution → `calibrate` → MATLAB →
publish) for 2026-01-30 against the real flight data, producing
`imap_mag_manual-norm-layer_20260130_v001.json` and publishing it to the datastore.
Full unit suite passes (956 + new tests); MATLAB is mocked in tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.pre-commit-config.yaml
```diff
@@ -30,11 +30,11 @@ repos:
...
-  - repo: https://github.com/igorshubovych/markdownlint-cli
-    rev: v0.49.0
-    hooks:
-      - id: markdownlint-fix
-        args: [--ignore, LICENSE.md]
+#   - repo: https://github.com/igorshubovych/markdownlint-cli
+#     rev: v0.48.0
+#     hooks:
+#       - id: markdownlint-fix
+#         args: [--ignore, LICENSE.md]
```
alastairtree: Uncommit this change, not relevant to this branch.

src/imap_mag/cli/calibrate.py
```diff
@@ -97,6 +104,21 @@ def calibrate(
...
     ] = SaveMode.LocalOnly,
+    metakernel: Annotated[
+        Path | None,
+        typer.Option(
+            help="Filename of the SPICE metakernel to use (scripted-l2 method). "
+            "Must exist in the spice/mk folder of the datastore. If omitted for "
+            "scripted-l2, one is generated.",
+        ),
+    ] = None,
+    matlab_repo_path: Annotated[
```
alastairtree: This level or one lower down the calls needs some validation so that matlab_repo_path must exist if the method is set to scripted L2 so that it is required

src/mag_toolkit/calibration/MatlabWrapper.py
```diff
@@ -33,19 +34,40 @@ def get_matlab_command():
...
+
+    Args:
+        command: The MATLAB command to run inside ``matlab -batch``.
+        first_call: Whether this is the first MATLAB call in the process (controls
+            whether the project path preamble is folded in).
+        timeout: Timeout in seconds for the MATLAB process.
+        cwd: Working directory to run MATLAB from. When calibrating with an
+            externally-acquired MATLAB project the working directory must be the
+            root of that project so its own ``addpath(pwd)`` logic resolves.
+        unset_display: If True, remove the ``DISPLAY`` env var for the MATLAB
```
alastairtree: Remove unset_display argument and instead always remove that var from the env when calling matlab

src/mag_toolkit/calibration/MatlabWrapper.py
```diff
@@ -33,19 +34,40 @@ def get_matlab_command():
...
+            process so it does not attempt to open plot windows.
+        include_project_paths: If True, allow prepending the imap-mag project
```
alastairtree: Merge the first_call and include_project_paths arguments into just the latter as they do the same thing

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+from mag_toolkit.calibration.CalibrationDefinitions import CalibrationMethod
+from mag_toolkit.calibration.MatlabWrapper import call_matlab
+
+from .CalibrationJob import CalibrationJob
+
+logger = logging.getLogger(__name__)
+
+# A single day of the MATLAB L2 calibration takes a few minutes; allow plenty of
+# headroom so slow days (e.g. cold MATLAB start plus SPICE furnishing) do not time out.
+SCRIPTED_L2_MATLAB_TIMEOUT_SECONDS = 60 * 30
```
alastairtree: This timeout value needs to flow from the general IMAP configuration file. It should have 2 versions of the timeout, based on normal or burst mode data, and the timeout should be based on per-day so calibration of 2 days of data has double the timeout. Normal mode should default to 10m per day in the config file. Burst mode should default to 60m, per day of data, timeout.

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+    "leapseconds",
+    "planetary_constants",
+    "science_frames",
+    "imap_frames",
+    "spacecraft_clock",
+    "attitude_history",
+    "pointing_attitude",
+    "planetary_ephemeris",
+    "ephemeris_reconstructed",
+]
```
alastairtree: Do not duplicate these values from the applicator, instead both flows should use the same set of values for kernel types and they should both be sourced from the configuration file.

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+    * resolves (or generates) the SPICE metakernel,
+    * invokes ``calibration.scripts.calibrate_l2_offsets`` from the root of the
+      acquired MATLAB repository,
+    * collects the produced layer JSON + CSV from the work folder.
+    """
+
+    def __init__(
+        self,
+        calibration_job_parameters: CalibrationJobParameters,
+        work_folder: Path,
```
alastairtree: Rather than passing in the work folder, pass in the application configuration object (which is built from the IMAP yaml file in the project root). You will need that to access the other refactoring siggion params like timeout and kernel types.

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+        metakernel: Path | str | None = None,
+    ):
+        super().__init__(calibration_job_parameters, work_folder)
+        self.name = CalibrationMethod.SCRIPTED_L2_CALIBRATION
+
+        if matlab_repo_path is None:
+            raise ValueError(
+                "A MATLAB calibration repository path is required for scripted L2 calibration."
+            )
+        self.matlab_repo_path = Path(matlab_repo_path)
```
alastairtree, line 56: Ensure this location actually exists and raise if not.
Also raise if the expected MATLAB script we are calling does not already exist as the MATLAB call will fail anyway.

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+        self.metakernel = metakernel
+
+    def _get_path_handlers(self, calibration_job_parameters: CalibrationJobParameters):
+        # The MATLAB script reads L1 science directly from the datastore, so there
+        # are no files to fetch into the work folder for this job.
+        return {}
```
alastairtree, line 90: I want this to work in 2 possible modes based on a new mode enum that is provided as an argument into the root calling calibrate flow. 
DATASTORE_ACCESS_MODE.READ_DIRECTLY is like the current code, assume MATLAB just reads from the datastore directly based on the path in the dynamic file config json. This is the default option and is used if the argument is not provided.
DATASTORE_ACCESS_MODE.LOCAL_WORK_FOLDER_COPY is a new mode where this function should prepare a local sparse version of the datastore folder into the work folder into a dynamically generated folder (and delete it afterwards). To achieve this the application config file will need to maintain a list of matching path patterns that represents all the various science and hk file types required by the MATLAB script and this preparation function should copy all the matching files into a new sparse datastore folder inside the work folder. It should only copy files that match the relevant days being calibrated or possibly the N days before or M days after which is specified in config. So the paths need generating dynamically based on the science content date being calibrated. Study the MATLAB script to understand if/which data from days before or after the day being calibrated is needed and set default config values. It should not try and copy in the entire datastore as that's massive. Study the existing MATLAB script to work out all the datastore file path patterns it needs as inputs. This local sparse datastore will also include the spice metakernel in the spice folder and all the kernels it references. It will also probably include the l1b or c science, the lo pivot platform HK, and various other HK files. Put this sparse datastore building logic into its own class. Ensure that there is enough disk space in the work folder before copying in all the files from the main datastore. 
All this work should happen in a way that is unified with the other calibration methods which also do some local copying.
When in LOCAL_WORK_FOLDER_COPY mode the path to the datastore in the json config file that is passed to MATLAB should point to the new sparse folder in the work folder.

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+                "ScriptedL2CalibrationJob requires a ScriptedL2CalibrationConfig, "
+                f"got {type(config).__name__}."
+            )
+
+        if not self._check_environment_is_setup():
+            raise FileNotFoundError(
+                "Environment has not been correctly set up for calibration."
+            )
+
+        if not self.matlab_repo_path.is_dir():
```
alastairtree: Check this earlier on in the process

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+            raise FileNotFoundError(
+                f"MATLAB calibration repository not found at {self.matlab_repo_path}."
+            )
+
+        date = self.calibration_job_parameters.date
+        matlab_mode = self.calibration_job_parameters.mode.value  # "norm" / "burst"
+        output_data_version = cal_handler.version
+
+        metakernel_filename = self._resolve_metakernel(date)
+        user_config_path = self._write_user_config()
```
alastairtree: Is this written into the work folder for the calibrate step? If not it should be.

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -0,0 +1,248 @@
...
+    def _resolve_metakernel(self, date: datetime) -> str:
+        """Return the metakernel filename to pass to MATLAB.
+
+        MATLAB looks the metakernel up under ``{sharepoint_flight_data}/spice/mk/``,
+        so only the bare filename is passed. If a metakernel was provided we verify
+        it exists there; otherwise we generate one (like the apply flow does) and
+        publish it to the datastore so MATLAB can find it.
+        """
+        assert self.data_store is not None
+        mk_dir = self.data_store / "spice" / "mk"
```
alastairtree: This magic strings patching logic should not live in here. The spice path handler is probably a better place

src/prefect_server/performCalibration.py
```diff
@@ -106,11 +193,36 @@ def calibrate_flow(
...
+
+    Args:
+        metakernel: Filename of the SPICE metakernel to use for the scripted-l2
+            method. Treated like ``spice_metakernel`` in ``apply_flow``: if provided
+            it must exist (in the datastore's spice/mk folder); if None and the
+            method is scripted-l2 one is generated.
+        matlab_repo: Where to acquire the MATLAB calibration code from for the
+            scripted-l2 method. A LocalFileSystem block (local path), a
+            GitHubRepository block (pulled into the work folder), the name of such
+            a block, or None for the other methods.
```
alastairtree: Needs documentation on all the args not just the new ones

src/prefect_server/performCalibration.py
```diff
@@ -106,11 +193,36 @@ def calibrate_flow(
...
+    """
+    matlab_repo_path: Path | None = None
+    if method == CalibrationMethod.SCRIPTED_L2_CALIBRATION:
+        app_settings = AppSettings()  # type: ignore
+        work_folder = app_settings.setup_work_folder_for_command(
+            app_settings.fetch_science
```
alastairtree: Add a new command config section in the app settings for calibration and use that so all calibration can have it's own application configuration and work folder. Add support for placeholders in the work folder name that is optional but allows the code to generate a unique work folder based on arguments to the flow. Calibrate work folders should be dynamically generated and include the content dates and the mode in the folder name based on the arguments to perform calibration flow.

tests/unit/test_performCalibration.py
```diff
@@ -116,3 +124,138 @@ def test_apply_flow_calls_apply(self):
             )
 
         mock_apply.assert_called_once()
+
+
+class TestGithubRepoName:
+    def test_ssh_url(self):
+        assert (
+            _github_repo_name(
+                "git@github.com:ImperialCollegeLondon/IMAP_MAG_Calibration.git"
```
alastairtree: Use a general URL here not so specific

imap-mag-config.yaml
```diff
@@ -59,6 +59,16 @@ plot_ialirt:
...
 
+# Configuration for the calibrate flow. work_sub_folder supports {date}/{mode}/{sensor}
+# placeholders so each calibrate run gets its own uniquely-named work folder.
+calibrate:
+    work_sub_folder: 'calibrate_{date}_{mode}'
+    scripted_l2_timeout_seconds_per_day_norm: 600    # 10 minutes per day of norm data
+    scripted_l2_timeout_seconds_per_day_burst: 3600  # 60 minutes per day of burst data
+    sparse_datastore:
+        days_before: 1
+        days_after: 1
```
alastairtree: These should be configurable on a per file path pattern basis because some big files like burst science you want to minimise the number of files you are copying. Not configured like this for all files.

imap-mag-config.yaml
```diff
@@ -59,6 +59,16 @@ plot_ialirt:
 calibration:
     work_sub_folder:
 
+# Configuration for the calibrate flow. work_sub_folder supports {date}/{mode}/{sensor}
+# placeholders so each calibrate run gets its own uniquely-named work folder.
+calibrate:
+    work_sub_folder: 'calibrate_{date}_{mode}'
+    scripted_l2_timeout_seconds_per_day_norm: 600    # 10 minutes per day of norm data
+    scripted_l2_timeout_seconds_per_day_burst: 3600  # 60 minutes per day of burst data
+    sparse_datastore:
```
alastairtree, line 93: I expect to see in here the list of file path patterns globs to build the sparse datastore, and for each pattern an optional number of days before and after. If before and after days not specified just copy the day being calibrated

src/imap_mag/config/CalibrationCommandConfig.py
```diff
@@ -0,0 +1,88 @@
...
+    "planetary_constants",
+    "science_frames",
+    "imap_frames",
+    "spacecraft_clock",
+    "attitude_history",
+    "pointing_attitude",
+    "planetary_ephemeris",
+    "ephemeris_reconstructed",
+]
+
```
alastairtree, line 19: Move this list of metakernel types into AppSettings yaml file and consumers should retrieve it from there instead

src/imap_mag/config/CalibrationCommandConfig.py
```diff
@@ -0,0 +1,88 @@
...
+    The SPICE metakernel and every kernel it references are always copied
+    separately (parsed from the metakernel), so they are not listed here.
+    """
+
+    days_before: int = 1
+    days_after: int = 1
+    per_day_patterns: list[str] = [
+        # L1 science for both sensors (level/mode filled per run; day = d).
+        "science/mag/{level}/{Y}/{m}/imap_mag_{level}_{mode}-mago_{Ymd}_v*.cdf",
```
alastairtree: Use yy and yyyy for 2 and 4 digit years instead of just Y. Prefer python style standard date time format string style substitution.

src/imap_mag/config/CalibrationCommandConfig.py
```diff
@@ -0,0 +1,88 @@
...
+        "calibration/inputs/Matrices/CalibrationMatricesV{matrix_version}.mat",
+        "calibration/inputs/Matrices/CalibrationMatricesV{matrix_version}.json",
+        # Interference / thruster profiles.
+        "calibration/inputs/Profiles/**/*",
+        # Pre-computed offsets (spin plane / spin axis / leinweber / kepko).
+        "calibration/calculated_offsets/**/*",
+        # Spin table + thruster activities (small, span the run).
+        "spice/spin/*",
+        "spice/activities/*",
+    ]
```
alastairtree: All patterns listed in this file should be instead live in The AppSettings yaml config file

src/imap_mag/config/CalibrationCommandConfig.py
```diff
@@ -0,0 +1,88 @@
...
+    # Folders that MATLAB references with different letter-casing than they are
+    # stored on the (case-insensitive) shared datastore. The MATLAB code reads
+    # thruster profiles via lowercase ``inputs/profiles`` but everything else via
+    # ``inputs/Profiles``; on the case-sensitive sparse copy we add a symlink for
+    # the lowercase spelling so both reads resolve.
+    case_insensitive_dir_aliases: list[str] = ["calibration/inputs/Profiles"]
```
alastairtree: If MATLAB has case sensitive folder access then we need to fix that code instead. Commit a fix in the calibration project but don't fix so profiles is always accessed in a case consistent correct way. We should not do any symlinking or file duplication to correct casing issues. Add a comment to the pr of the places where MATLAB code is case sensitive as well.
Remove the case_insensitive_dir_aliases logic.

alastairtree:
Re: MATLAB case-sensitive folder access — removed the `case_insensitive_dir_aliases` symlink workaround and fixed the casing at the source in the calibration project instead.

**Where the MATLAB code was case-inconsistent** (`IMAP_MAG_Calibration`):
- `+calibration/+configuration/ThrusterConfiguration.m` built the thruster profile path with a lowercase `fullfile(calibration_dir.location, 'inputs', 'profiles', ...)`, whereas every other consumer resolves it via `CalibrationDir.getProfilesFolder()` → `fullfile(this.location, "inputs", "Profiles")` (capital `Profiles`, matching the on-disk folder). On a case-sensitive filesystem the lowercase `profiles` read fails.
  - `+calibration/+configuration/InterferenceConfiguration.m` and `PreThrusterConfiguration.m` already use `getProfilesFolder()` (correct).

**Fix** (committed on the calibration repo's `feat/pipelinification` branch): `ThrusterConfiguration.m` now uses `calibration_dir.getProfilesFolder()`, so all profile reads use the canonical `Profiles` casing consistently. That commit needs pushing to the calibration repo (it was made locally in the dev container).

src/imap_mag/io/FileFinder.py
```diff
@@ -255,6 +256,143 @@ def find_latest_science_by_date(
...
+
+    def find_dated_files_with_fallback(
+        self,
+        relative_pattern: str,
+        start_date: datetime,
+        end_date: datetime,
+        days_before: int = 0,
+        days_after: int = 0,
+        get_previous_if_empty: bool = False,
+        max_lookback_days: int = 3650,
```
alastairtree: This should always be gated by IMAP mission start date constant not a max days

src/mag_toolkit/calibration/SparseDatastoreBuilder.py
```diff
@@ -0,0 +1,182 @@
...
+                )
+            else:
+                matches = self._finder.find_dated_files_with_fallback(
+                    named,
+                    start_date=search_start,
+                    end_date=search_end,
+                    days_before=pattern.days_before,
+                    days_after=pattern.days_after,
+                    get_previous_if_empty=pattern.get_previous_if_empty,
+                )
```
alastairtree: It would be much cleaner to have one file matching method which supports all the possible searching options, so that for example highest sequence matching and getting previous if empty options can be combined. Ultrathink on this problem to come up a better reusable function that can match and copy files dynamically based on all the configured options

src/mag_toolkit/calibration/SparseDatastoreBuilder.py
```diff
@@ -0,0 +1,182 @@
...
+    SPICE metakernel and exactly the kernels it references are always copied
+    separately, with the metakernel's ``PATH_VALUES`` normalised to the relative
+    ``spice`` folder so it furnishes from the sparse root.
+    """
+
+    def __init__(
+        self,
+        source_datastore: Path,
+        config: SparseDatastoreConfig,
+        disk_usage_threshold: float = 0.95,
```
alastairtree: Threshold must flow from yml configuration file in app settings and not rely on a default in code

src/imap_mag/io/FileFinder.py
```diff
@@ -255,6 +256,143 @@ def find_latest_science_by_date(
             f"No science file found for date {date.strftime('%Y-%m-%d')} and mode {mode.value}"
         )
 
+    _COVERAGE_PLACEHOLDER_RE = re.compile(r"\{from_doy\}|\{to_doy\}|\{sequence\}")
+
+    def find_by_coverage_window(
```
alastairtree: Merge the 2 new find methods in this file into one so that any combination of the arguments can work

src/imap_mag/io/FileFinder.py
```diff
@@ -255,6 +256,143 @@ def find_latest_science_by_date(
...
+
+        best: dict[tuple[datetime, datetime], tuple[Path, int]] = {}
+        for path, window_start, window_end, sequence in overlapping:
+            key = (window_start, window_end)
+            if key not in best or sequence > best[key][1]:
+                best[key] = (path, sequence)
+        return sorted(path for path, _ in best.values())
+
+    @staticmethod
+    def _coverage_window_regex(pattern: str) -> re.Pattern:
```
alastairtree, line 412: Add a check on the pattern for any types of {substitutions} that we don't support and give a helpful error if the pattern includes any that are not supported

src/mag_toolkit/calibration/SparseDatastoreBuilder.py
```diff
@@ -0,0 +1,182 @@
...
+                    days_after=pattern.days_after,
+                    get_previous_if_empty=pattern.get_previous_if_empty,
+                )
+
+            for source in matches:
+                relative = source.relative_to(self.source_datastore)
+                copied += self._copy_file(source, target_root / relative)
+
+        copied += self._copy_metakernel_and_kernels(metakernel_filename, target_root)
+
```
alastairtree, line 105: Add a space check inside this loop. Also count the name and size of each file and log each file being copied, and the total size at the end. Suggest both put inside the copy file method?

src/mag_toolkit/calibration/SparseDatastoreBuilder.py
```diff
@@ -0,0 +1,182 @@
...
+        # spice_metakernal_root before furnishing). A relative value also avoids
+        # SPICE's limit on the length of a metakernel path token.
+        dest_mk = SPICEPathHandler.get_metakernel_path(target_root, metakernel_filename)
+        dest_mk.parent.mkdir(parents=True, exist_ok=True)
+        dest_mk.write_text(self._rewrite_metakernel_path_values(source_mk.read_text()))
+        count += 1
+        return count
+
+    @staticmethod
+    def _parse_metakernel_kernels(metakernel_path: Path) -> list[str]:
```
alastairtree: We have other classes for metakernel handling I think, might be a better place to put this code?

alastairtree:
Ok this beast is now in a decent working state and has been tested locally (against my branch in imap cal) and can indeed calibrate some normal mode data correctly and then put the layer files into the datastore. Also it has all of mine and @mhairifin 's wish list:
- copy local work folder
- most recent hk that has data when gap
- pull hk with day boundary extensions
- deduplicate layer files based on the data in the csv (even though matlab does not put the hashes into the layer files)
- one step cal+apply
- An improved config ui that accepts all the args based on the type of cal you are doing

PR #331, opened 2026-07-15
Add NOAA fetch and cli

### LINKED ISSUE #340: Fetch NOAA's data and save it to CSV files
Getting the data from the client and creating one file per day/spacecraft/instrument combination. A CLI should enable testing the fetcher locally. The downloaded data should be processed so only the relevant fields provided by the end points are saved.

dalonsoa:
Adds NOAA fetch class and CLI using it to pull SOLAR-1 and ACE data and save it as CSV files.

Largely copies what IALiRT does and, indeed, I feel there's some duplication that could be removed with some abstraction.

This is the second step towards ingesting the SOLAR-1 and ACE data in the database.

### TODO
- [x] Add tests


dalonsoa:
@alastairtree tests are missing, but functionality-wise this should be complete. I'll add them as soon as possible, but probably not this week. Let me know if you spot anything plainly wrong or that needs improvement.

dalonsoa:
@alastairtree I think this is ready. What's next, I'm not sure. Is it uploading data to the DB? Configuring the prefect workflow? Any hints on what files to look at would be useful.

alastairtree:
Had a quick skim and looks ok Assume you have tested it actually works as a cli to download them? I would probably add a wiremock based test so you can actually verify that it does download data from a real api in your tests.

Then I think you probably want to have a look at poll_ialirt_flow in main to see how we set these up as flows and use the pipeline/stage classes so that it can run in 2 modes - download a date range, or automatic run where it maintains a datetime of current progress in the database and uses that to crawl forwards in time adding just the new data.

dalonsoa:
> Had a quick skim and looks ok Assume you have tested it actually works as a cli to download them?

Yes, it works well from the CLI.

> I would probably add a wiremock based test so you can actually verify that it does download data from a real api in your tests.

Will do it right now. Then this is ready to merge.

> Then I think you probably want to have a look at poll_ialirt_flow in main to see how we set these up as flows and use the pipeline/stage classes so that it can run in 2 modes - download a date range, or automatic run where it maintains a datetime of current progress in the database and uses that to crawl forwards in time adding just the new data.

That's https://github.com/ImperialCollegeLondon/imap-pipeline-core/issues/342 . Will work on it today. @miruuna already gave me some hints on how to do it.

dalonsoa:
Integration test using wiremock added. I think this should be ready to merge.

src/imap_mag/cli/fetch/noaa.py
```diff
@@ -0,0 +1,98 @@
...
+    spacecraft: Annotated[
+        Literal["SOLAR1", "ACE"],
+        typer.Option(
+            help="Spacecraft to download data for. Must be 'SOLAR1' or 'ACE'",
+        ),
+    ],
+    instrument: Annotated[
+        Literal["mag", "wind"],
+        typer.Option(
+            help="Instrument data to download. Must be 'mag' or 'palsma'",
```
alastairtree, line 71:
```suggestion
            help="Instrument data to download. Must be 'mag' for the magnetic field instrument or 'wind' for the plasma instrument",
```

src/imap_mag/cli/fetch/noaa.py
```diff
@@ -0,0 +1,98 @@
...
+
+def _create_fetch_noaa(app_settings: AppSettings) -> FetchNOAA:
+    """Create a FetchNOAA instance with common configuration."""
+
+    data_access = NOAARTSWApiClient(
+        app_settings.fetch_solar1_ace.api.url_base,
+    )
+    datastore_finder = FileFinder(app_settings.data_store)
+    work_folder = app_settings.setup_work_folder_for_command(
+        app_settings.fetch_solar1_ace
```
alastairtree, line 25: Not sure if these ever run in parallel but would always be using the same folder for all craft/instrument so potential for one run stepping on the data files of another run. Can pass a `name_context` to make it unique if this is a possibility?

src/imap_mag/download/FetchNOAA.py
```diff
@@ -0,0 +1,210 @@
...
+                "Invalid spacecraft requested. "
+                f"It must be 'SOLAR1' or 'ACE', but '{spacecraft}' found"
+            )
+
+        if instrument not in ("mag", "wind"):
+            raise ValueError(
+                f"Invalid instrument type requested for {spacecraft}. "
+                f"It must be 'mag' or 'wind', but '{instrument}' found"
+            )
+
```
alastairtree, line 62: I think this logic is duplicated with some in `NOAAApiClient` and should probably only live in one place so perhaps delete these and leave it in the client?

src/imap_mag/download/FetchNOAA.py
```diff
@@ -0,0 +1,210 @@
...
+            instrument=instrument,
+        )
+
+        if not downloaded:
+            logger.debug(
+                f"No {spacecraft} {instrument} data downloaded from NOAA RTSW."
+            )
+            return dict()
+
+        process_fn = _process_noaa_mag if instrument == "mag" else _process_noaa_wind
```
alastairtree, line 74: Minor thing - suggest change this to a switch statement that raises if not a known instrument to guard against adding a new instrument later that silently uses _process_noaa_wind

src/imap_mag/download/FetchNOAA.py
```diff
@@ -0,0 +1,210 @@
...
+        if instrument not in ("mag", "wind"):
+            raise ValueError(
+                f"Invalid instrument type requested for {spacecraft}. "
+                f"It must be 'mag' or 'wind', but '{instrument}' found"
+            )
+
+        downloaded: list[dict[str, Any]] = self._data_access.get_data(
+            spacecraft=spacecraft,
+            instrument=instrument,
+        )
```
alastairtree, line 66: `list[dict[str, Any]]` is a bit ambiguous what this data contains. Might be worth a comment to give a sample or description

src/imap_mag/download/FetchNOAA.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+            return dict()
+
+        process_fn = _process_noaa_mag if instrument == "mag" else _process_noaa_wind
+        downloaded_data = process_fn(pd.DataFrame(downloaded))
+        return self._add_to_files(spacecraft, instrument, downloaded_data)
+
+    def _add_to_files(
+        self,
+        spacecraft: Literal["SOLAR1", "ACE"],
```
alastairtree, line 80: `Literal["SOLAR1", "ACE"]` and `Literal["mag", "wind"]` are defined multiple times in this class. Make them an emun or something to remove duplication?

src/imap_mag/download/FetchNOAA.py
```diff
@@ -0,0 +1,210 @@
...
+
+        for day_info, daily_data in data.groupby(dates):
+            date: datetime = day_info[0] if isinstance(day_info, tuple) else day_info  # type: ignore
+
+            daily_dates = self._get_index_as_datetime(daily_data)
+            min_daily_date = min(daily_dates)
+            max_daily_date = max(daily_dates)
+
+            path_handler = NOAAPathHandler(
+                mission=spacecraft, instrument=instrument, content_date=max_daily_date
```
alastairtree, line 103: Using content_date to not just hold the date of the content, but also hold the datetime of the most recent data entry (and later using that to update the workflow progress table) is pattern I think introduced and wished i had not. It looks like it can safely truncated to just a date but that means potentially re-downloading all day. I have been burned by it before and had to fix. Better to add an explicit `max-record-date` to the handler I think - https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/main/src/imap_mag/io/file/IALiRTPathHandler.py#L26

src/imap_mag/io/file/NOAAPathHandler.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def get_folder_structure(self) -> str:
+        super()._check_property_values("folder structure", ["content_date"])
+        assert self.content_date
+
+        return (Path(self.root_folder) / self.content_date.strftime("%Y/%m")).as_posix()
+
+    def get_filename(self) -> str:
+        super()._check_property_values("file name", ["content_date"])
+        assert self.content_date
```
alastairtree, line 36: This should probably error if mission or instrument is not defined, in addition to `content_date`

tests/datastore/noaa/rtsw_mag_1m.json
alastairtree, line 1: These two data files are quite chunky. Might be tempted to trim them to just a few records unless there is a need for a full payload?

tests/integration/test_fetchNOAA.py
```diff
@@ -0,0 +1,73 @@
...
+    instrument,
+    wiremock_manager,
+    capture_cli_logs,
+) -> None:
+    # Set up.
+    filename = f"rtsw_{instrument}_1m.json"
+    mock_data_path = NOAA_DATA_PATH / filename
+    with open(mock_data_path) as f:
+        response = json.load(f)
+    wiremock_manager.add_string_mapping(
```
alastairtree, line 40: You can use `wiremock_manager.add_file_mapping` to map a file more directly and save yourself the read and load.

tests/integration/test_fetchNOAA.py
```diff
@@ -0,0 +1,73 @@
...
+        f"/{filename}",
+        json.dumps(response),
+    )
+
+    # Create FetchNOAA instance.
+    work_folder = Path(tempfile.mkdtemp())
```
alastairtree, line 46: We do have a fixture `dynamic_work_folder` to build a dynamic work folder for you (so you can just use default AppSettings) and it gets cleaned up after the tests for you, or even just fixture `temp_folder_path` if you want it

tests/integration/test_fetchNOAA.py
```diff
@@ -0,0 +1,73 @@
...
+            f"{spacecraft}_{instrument}_noaa_{file_path.stem.split('_')[-1]}.csv"
+        )
+        assert file_path.name == expected_filename
+
+        date_str = file_path.stem.split("_")[-1]
+        year = date_str[:4]
+        month = date_str[4:6]
+        expected_data = pd.read_csv(NOAA_DATA_PATH / year / month / expected_filename)
+        actual_data = pd.read_csv(file_path)
+        pd.testing.assert_frame_equal(expected_data, actual_data)
```
alastairtree, line 73: We have been keeping these sort of expected file contents (as opposed to test input data) in `tests/test_truth`

tests/unit/test_fetchNOAA.py
alastairtree, line 1: Nice tests!

alastairtree (approved):
Lots of minor nitpicks in here - sorry! - but looks good, thank you. Merge this and of the other approved ones when you are happy and sorry it took ages to get this reviewed properly.

src/imap_mag/download/FetchNOAA.py
alastairtree: Also I found a bunch of subtle bugs in the ialirt pipeline in this section to do with it running at different times of the day and updating / overwriting files so worth comparing to the latest version.

src/imap_mag/io/file/NOAAPathHandler.py
dalonsoa: `mission` and `instrument` are both mandatory arguments, so they need to be defined always. An instance of the class cannot be created without them.

tests/datastore/noaa/rtsw_mag_1m.json
dalonsoa: We need enough input data to produced processed CSV covering 2 days, so we can test all functionality. I'll try to shorten them a bit.

src/imap_mag/download/FetchNOAA.py
dalonsoa: This pipeline is a bit different in the sense that it always downloads the last 24h of data, regardless of when was the last available date. The `content_date` is just used to define the folder in which to save the stuff, to to decide what to download, so I think we are safe here. Either that or I'm not understanding the risk.

About the other changes, again I do not think they affect this particular pipeline, but good to know about then, just in case.

src/imap_mag/cli/fetch/noaa.py
dalonsoa: Not sure what you mean. I pretty much followed `ialirt` code, which does not include any `name_context`. What's it purpose?

dalonsoa:
@alastairtree I've addressed most of your comments, and just left a couple without action and with some questions from my side. I'm merging this now, but we can address those if needed in the future.

PR #344, opened 2026-07-22
341 upload noaa

### LINKED ISSUE #341: Upload data to database
The content of the CSV files created in #340 should be uploaded to the database. Relevant fields and tables will need to be created. Probably we want one table per spacecraft/instrument combination (so 4 tables, at the moment), but I wonder if it would be more logical and convenient to have one table per spacecraft. Whatever is easier with the current infrastructure and existing functionality will do.

dalonsoa:
Updates the configuration to use the existing workflows to ingest NOAA csv files and to delete them after 7 days. `crump` was used to update the configuration in `imap-db-ingest-config.yaml`.

This PR only deals with files, NOT the database.

Close #341

### DIFF
--- a/imap-db-ingest-config.yaml
+++ b/imap-db-ingest-config.yaml
@@ -1,7 +1,6 @@
 jobs:
   imap_mag_l1_hsk-procstat_v:
     target_table: hsk_procstat
-    filename_match: imap_mag_l1_hsk-procstat_*_v*.csv
     id_mapping:
       shcoarse:
         db_column: shcoarse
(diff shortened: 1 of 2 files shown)


alastairtree (approved):

alastairtree:
Looks good thanks. Even If the CSV files are deleted the data is still in the database so yeah i guess we need a flow to delete those

dalonsoa:
OK, I'll make that a separate pipeline.

alastairtree:
Happy for you to merge whenever you like - i don’t want to merge your stack in the wrong order.

PR #346, opened 2026-07-23
Add NOAA pipeline

### LINKED ISSUE #342: Automate pulling data from NOAA and saving it in the database
The steps in #340 and #341 should all be automated so they run every 5 min for the 4 combinations of spacecraft/instrument.

dalonsoa:
Mostly copies the iALrt one, just changing the downloader and re-using the other stages.

As the NOAA fetcher always gets the last 24h of data (what the json file contains), I'm not entirely sure if we actually need the `GetProcessingDatesStage`. We don't do any filtering of the downloaded data, after all.

Close #342

### DIFF
--- a/src/imap_mag/data_pipelines/DownloadNOAAStage.py
+++ b/src/imap_mag/data_pipelines/DownloadNOAAStage.py
@@ -1,0 +1,48 @@
+from imap_mag.data_pipelines import PROGRESS_DATE_CONTEXT_KEY, FileRecord, Record, Stage
+from imap_mag.download.FetchNOAA import FetchNOAA
+
+
+class DownloadNOAAStage(Stage):
+    """Download NOAA RTSW data for a spacecraft/instrument and emit one FileRecord per downloaded day."""
+
+    def __init__(
+        self,
+        spacecraft: str,
+        instrument: str,
+        fetcher: FetchNOAA,
+    ):
+        """Initialise the stage.
+
+        Args:
+            spacecraft: Spacecraft to download data for. Must be 'SOLAR1' or 'ACE'.
+            instrument: Instrument to download data for. Must be 'mag' or 'plasma'.
+            fetcher: FetchNOAA instance used to perform the download.
+        """
+        super().__init__()
+        self.spacecraft = spacecraft
+        self.instrument = instrument
+        self.fetcher = fetcher
+
+    async def process(self, item: Record, context: dict, **kwargs):
+        self.logger.info(f"Downloading NOAA {self.spacecraft} {self.instrument} data.")
+
+        downloaded = self.fetcher.download_csv(
+            spacecraft=self.spacecraft,  # type: ignore
+            instrument=self.instrument,  # type: ignore
+        )
+
+        if not downloaded:
+            self.logger.info(
+                f"No NOAA {self.spacecraft} {self.instrument} data downloaded."
+            )
+            return
+
+        for file_path, path_handler in downloaded.items():
+            content_date = path_handler.content_date  # type: ignore
+            context[PROGRESS_DATE_CONTEXT_KEY] = content_date
+
+            await self.publish_next(
+                FileRecord(file_path, content_date),
+                context,
+                **kwargs,
+            )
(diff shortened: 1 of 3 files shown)


src/imap_mag/data_pipelines/DownloadNOAAStage.py
```diff
@@ -0,0 +1,48 @@
...
+        self.spacecraft = spacecraft
+        self.instrument = instrument
+        self.fetcher = fetcher
+
+    async def process(self, item: Record, context: dict, **kwargs):
+        self.logger.info(f"Downloading NOAA {self.spacecraft} {self.instrument} data.")
+
+        downloaded = self.fetcher.download_csv(
+            spacecraft=self.spacecraft,  # type: ignore
+            instrument=self.instrument,  # type: ignore
```
alastairtree, line 31: Are these type warning ignore comments because of the stings to literal conversion? How annoying!

src/imap_mag/data_pipelines/NOAAPipeline.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self._fetcher = FetchNOAA(
+            data_access=self._client,
+            work_folder=work_folder,
+            datastore_finder=datastore_finder,
+        )
+
+        self._datetime_provider = datetime_provider
+
+    def build(self, run_params: AutomaticRunParameters | FetchByDatesRunParameters):  # type: ignore
```
alastairtree, line 58: There is no way to pass date filters to these APIs i think - they always and only return the last 24hrs data? If so the FetchByDatesRunParameters can be removed I think, it just adds confusion.

alastairtree (changes requested):
Thanks for this. This API is different to any we have integrated with before in that it does not allow any date based filtering I think? In that case we must make a request at minimum every 24hrs otherwise we lose data. We can manage that operationally but please add the noaa examples to the readme and add a comment somewhere that explains the PAI just returns the last 24hrs data from now backwards.

A couple of requests:
- [ ] I think remove the fetch by dates params - they are just confusing here.
- [ ] And so then i think the whole GetProcessingDatesStage is not needed? Just always get the last 24hrs?
- [ ] Are you intending to add a prefect flow that actually calls the pipelines on a schedule (or is that another PR)? I think the prefect flows should be in this PR so we can actually run the pipelines and see the results and run this code in the test env for a day or so to see what happens. 
- [ ] Some of our other CLIs just run/use the pipeline class as well, rather than having the similar pipeliney logic in the cli. Ideally the CLI should just be a nice way to run the same code without needing a prefect server but have minimal specific logic. (IALiRT was a bit weird in that the CLI and the flow do not share/run the pipeline class, i think that is probably a tech debt). So I propose the CLI and the Flow both call NOAAPipeline?
- [ ] Call each craft/instrument pipeline one after the other? Nice if each was a prefect task for UI visibility/log filtering, and a failure in one did not stop the others.

How is the date in the filename related to the timestamps of the data in the file? All our other science with a day in the filename has only data on that day in it. I think because you are grouping by date this ends up the same updating today and yesterdays files, and each day file only contains data in that day? If so then great, I think that is right. The datastore is smart and will not update/republish yesterdays file once it has a full dataset because the file hash will be the same. That way you are usually only updating today's file, even if you are downloading the last 24hrs.

Thank you, and sorry for the slow PR feedback.

src/imap_mag/data_pipelines/DownloadNOAAStage.py
dalonsoa: Yep. Super annoying.

src/imap_mag/data_pipelines/NOAAPipeline.py
dalonsoa: Yes, that was my thought, as well.

dalonsoa:
No idea why the checks are not running...

alastairtree (approved):
Great - thanks. Please can you run all 4 pipelines versions and the postgres import in test env so database gets updated with all the new table schema so i can use that to migrate the prod db schema in azure

dalonsoa:
Done! Are there more meaningful log messages when submitting stuff to the DB, now?

dalonsoa:
@alastairtree , is there anything blocking merging this and deploying in production? I'm not in a rush, but I can do very little adjusting the web until this is sorted.

alastairtree:
No there is not, sorry. I've just been off for a couple of days. The release to prod is a bit of a faff because the replicated database tables need setting up before I can release so I just need to do that because it's a painfully manual step currently! Ideally we would have the cloud instance under accepted continuous deploy but it's a bit fiddly. Reckon I'll get there this afternoon

PR #350, opened 2026-07-30
feat: publish calculated spin-plane offsets to the datastore

mhairifin:
Integrate the scripted-L2 MATLAB calibration's optional offsets output with the datastore.

- Add a `write_offsets` option to ScriptedL2CalibrationConfig so it can be toggled from the Prefect UI. When set, MATLAB is told to write the calculated spin-plane offsets into an `offsets/` folder in the run's work folder.
- Add CalculatedOffsetsPathHandler mapping to calibration/calculated_offsets/{spin_plane,spin_optimised}. The two products share a file name and are distinguished by their folder.
- After calibration, publish each collected offsets CSV via the datastore file manager, which compares content by hash against the latest datastore version and upversions (and indexes in the database) only when it differs.

Tests cover the new path handler, the flag/config/user-config threading and CSV collection, the CLI end-to-end publish/upversion/reuse behaviour, and database indexing of the offsets files.

This also exposed an architecture issue where the base spin plane offsets and the optimised offsets files had the same name, so I have made small accompanying changes in the cal code - https://github.com/ImperialCollegeLondon/IMAP_MAG_Calibration/pull/78/changes - so that the offsets file names are unique.

### DIFF
--- a/src/imap_mag/cli/calibrate.py
+++ b/src/imap_mag/cli/calibrate.py
@@ -15,7 +15,10 @@ from imap_mag.config import (
 )
 from imap_mag.db.Database import Database
 from imap_mag.io import DatastoreFileManager, FileFinder
-from imap_mag.io.file import CalibrationLayerPathHandler
+from imap_mag.io.file import (
+    CalculatedOffsetsPathHandler,
+    CalibrationLayerPathHandler,
+)
 from imap_mag.util import ScienceMode
 from mag_toolkit.calibration import (
     CalibrationJob,
@@ -276,8 +279,46 @@ def _calibrate_for_date(
         data_path, path_handler=calibration_handler.get_equivalent_data_handler()
     )
 
+    # Publish any calculated spin-plane offsets the calibration wrote to the work
+    # folder, comparing against the datastore and upversioning when they differ.
+    _publish_calculated_offsets(
+        getattr(calibrator, "written_offset_files", []),
+        outputManager,
+        FileFinder(settings_for_output.data_store),
+    )
+
     logger.info(
         f"Calibration complete for {start_date.strftime('%Y-%m-%d')} ({mode.value}) written to {output_calibration_path!s}."
     )
 
     return output_calibration_path
+
+
+def _publish_calculated_offsets(
+    offset_files: list[Path],
+    output_manager,
+    offsets_finder: FileFinder,
+) -> None:
+    """Publish calculated spin-plane offset CSVs to the datastore.
+
+    Each file is versioned relative to the latest existing offsets for the same
+    sensor/date/offset-type: ``add_file`` reuses that version when the content is
+    identical (hash match) and upversions when it differs, indexing the result in
+    the database when the output manager is database-backed.
+    """
+    for offset_file in offset_files:
+        offset_handler = CalculatedOffsetsPathHandler.from_work_folder_file(offset_file)
+
+        latest = offsets_finder.find_latest_version_by_handler(
+            offset_handler, throw_if_not_found=False
+        )
+        if latest is not None:
+            existing = CalculatedOffsetsPathHandler.from_filename(latest.name)
+            offset_handler.version = existing.version if existing else 1
+        else:
+            offset_handler.version = 1
+
+        (published_offset_path, _) = output_manager.add_file(
+            offset_file, path_handler=offset_handler
+        )
+        logger.info(f"Published calculated offsets to {published_offset_path!s}.")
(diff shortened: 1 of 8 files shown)


src/imap_mag/cli/calibrate.py
```diff
@@ -276,8 +279,46 @@ def _calibrate_for_date(
         data_path, path_handler=calibration_handler.get_equivalent_data_handler()
     )
 
+    # Publish any calculated spin-plane offsets the calibration wrote to the work
+    # folder, comparing against the datastore and upversioning when they differ.
+    _publish_calculated_offsets(
+        getattr(calibrator, "written_offset_files", []),
+        outputManager,
+        FileFinder(settings_for_output.data_store),
+    )
```
alastairtree, line 288: I think this is a bit of a hack  - only the "manual" (which we should rename!) matlab cal produces these so it feels a bit wrong here. I think we should fix this properly - suggest we refactor so every calibration can return a list of any number of files and then we let the path handlers understand what the files are how to store them? Then we dont need to do anything spectial for these offsets except add a path handler for them, and open it up to be able to write multiple layers in matlab at once? Maybe we would need to just pass the version from python and then let it assume assume all files should be generated in the same version?

src/imap_mag/io/file/CalculatedOffsetsPathHandler.py
```diff
@@ -0,0 +1,149 @@
...
+        return cls(
+            sensor=match["sensor"],
+            offset_type=_OFFSET_TYPE_BY_DESCRIPTOR[match["descr"]],
+            content_date=datetime.strptime(match["date"], "%Y%m%d"),
+            version=int(match["version"]),
+            extension=match["ext"],
+        )
+
+    @classmethod
+    def from_work_folder_file(cls, file: Path) -> "CalculatedOffsetsPathHandler":
```
alastairtree, line 124: Why not just put this logic in the main from filename? Always check the parent folder - since they should always be in a properly named folder by requirement?

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -31,6 +32,11 @@
 # DatastoreAccessMode.LOCAL_WORK_FOLDER_COPY.
 SPARSE_DATASTORE_FOLDER_NAME = "sparse_datastore"
 
+# Folder (inside the work folder) MATLAB writes the calculated spin-plane offsets to
+# when write_offsets is enabled. Mirrors the calibration/calculated_offsets datastore
+# layout: one CSV per sensor per day under each OFFSET_TYPES sub-folder.
+OFFSETS_FOLDER_NAME = "offsets"
```
alastairtree, line 38: Propose this be `calculated_offsets` and then they are always arranged in `calculated_offsets\<type>` and this path name can live in one place in the file path handler?

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -130,6 +139,13 @@ def run_calibration(
         user_config_path = self._write_user_config(matlab_datastore)
         self._paths_needing_cleanup.append(user_config_path)
 
+        # MATLAB writes offsets into pre-existing sub-folders, so create them upfront.
+        if config.write_offsets:
+            for offset_type in OFFSET_TYPES:
+                (self.work_folder / OFFSETS_FOLDER_NAME / offset_type).mkdir(
+                    parents=True, exist_ok=True
+                )
```
alastairtree, line 147: Do we need to do this? MATLAB will create the folder if it needs to?

alastairtree (changes requested):

alastairtree:
@mhairifin I was thinking about this and think it would be better if the python calling code allowed the MATLAB to return any number of files as long as they have a handler. That way it is open to extension without needing to rewrite the python interface each time you want to add more files to the datastore and you have the option of returning a set of layers rather than just one. In that end I have created #355 and https://github.com/ImperialCollegeLondon/IMAP_MAG_Calibration/pull/82

If you agree with this new plan then I think this PR gets closed and we rewrite it on top of #355?

mhairifin:
> mhairifin I was thinking about this and think it would be better if the python calling code allowed the MATLAB to return any number of files as long as they have a handler. That way it is open to extension without needing to rewrite the python interface each time you want to add more files to the datastore and you have the option of returning a set of layers rather than just one. In that end I have created #355 and [ImperialCollegeLondon/IMAP_MAG_Calibration#82](https://github.com/ImperialCollegeLondon/IMAP_MAG_Calibration/pull/82)
> 
> If you agree with this new plan then I think this PR gets closed and we rewrite it on top of #355?

Thanks Alastair, I agree that sounds like a more flexible approach that should save us from having this issue in the future. I will take a look at your other PR

PR #352, opened 2026-07-31
fix: assign max+1 minor version instead of first available slot

alastairtree:
`DBIndexedDatastoreFileManager.__get_next_available_version` searched for the **first available gap** in existing version numbers rather than `max + 1`. For example, if a layer at v002 was the only active record backfilling the gap instead of advancing beyond the highest known version.

## Tests

- Added `test_DBIndexedDatastoreFileManager_get_next_available_version_uses_max_plus_one_not_first_gap`: verifies that when only v002 is active in the DB, the next assigned version is v003, not v001.
- All existing unit and mock-based integration tests continue to pass.


mhairifin (approved):
Looks good to me! Thanks for taking a look at this so fast

PR #354, opened 2026-08-01
fix: use xvfb-run to provide virtual display for headless MATLAB plotting

alastairtree:
## Problem

When the L2 scripted calibration flow runs MATLAB in the Docker container, MATLAB crashes while writing a report file that contains plots. The root cause is that MATLAB's graphics subsystem attempts to initialise X11 display infrastructure even in `-batch` mode when rendering figures — and the container has no display server.

Previously the code worked around this by unsetting `DISPLAY` before spawning MATLAB, but that only prevents MATLAB from trying to connect to an *existing* display; it does not suppress the underlying graphics initialisation that fails when no display is available at all.

## Fix

Two changes:

**`deploy/MATLAB-Dockerfile`** — add `xvfb` (X Virtual Framebuffer) to the apt packages installed in the runtime image. This provides the `xvfb-run` helper without requiring a desktop environment.

**`src/mag_toolkit/calibration/MatlabWrapper.py`** — introduce `_get_display_wrapper()`, which returns `["xvfb-run", "--auto-servernum"]` when `xvfb-run` is on the PATH (Linux only), or an empty list otherwise. The MATLAB invocation is prefixed with this wrapper so each call gets its own ephemeral virtual framebuffer. `xvfb-run` sets its own isolated `DISPLAY` for the MATLAB child process, so MATLAB can create and save figures off-screen.

The graceful fallback (empty wrapper) means developer machines without `xvfb` installed continue to work as before — the wrapper is purely additive in the Docker environment.

`DISPLAY` is still removed from the environment dictionary passed to `subprocess.Popen`, so the host display can never leak through.

Docs: https://uk.mathworks.com/help/slcheck/padv/ug/tips-for-setting-up-ci-agents.html

### DIFF
--- a/deploy/MATLAB-Dockerfile
+++ b/deploy/MATLAB-Dockerfile
@@ -105,6 +105,7 @@ RUN apt-get update && apt-get install -y \
     libcups2t64 \
     libncurses6 \
     libasound2t64 \
+    xvfb \
     git \
     wget \
     && apt-get clean && \
--- a/src/mag_toolkit/calibration/MatlabWrapper.py
+++ b/src/mag_toolkit/calibration/MatlabWrapper.py
@@ -79,6 +79,21 @@ def get_matlab_command():
         return "matlab"
 
 
+def _get_display_wrapper() -> list[str]:
+    """Return an xvfb-run prefix if available, otherwise an empty list.
+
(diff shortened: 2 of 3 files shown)


alastairtree:
This should fix your display plots/reports not working @mhairifin

mhairifin (approved):
Looks good!

PR #355, opened 2026-08-01
feat: calibration multi-file output + version-number override (v6.3.0)

alastairtree:
## Summary

- **Multi-file output loop**: `CalibrationJob.run_calibration` now returns `Sequence[Path]` instead of a fixed `tuple[Path, Path]`. The save loop in `calibrate._calibrate_for_date` iterates over every returned path and resolves a handler via `FilePathHandlerSelector.find_by_path()`, publishing all files MATLAB writes — not just the standard JSON+CSV pair.
- **`output_version` passed to MATLAB**: `ScriptedL2Calibration` appends `output_version=[major minor]` to the MATLAB command so MATLAB owns filename construction. Python collects whatever files appear in the dedicated `outputs/` subfolder after the run.
- **`--version-number-override MAJOR MINOR` CLI option**: Forces a specific `(major, minor)` version instead of auto-incrementing. `_validate_version_number_override` rejects bools, non-integers, negatives, and out-of-range values. When set, `allow_overwrite=True` is propagated through `VersionedPathHandler`, `CalibrationLayerPathHandler`, `DatastoreFileManager`, and `DBIndexedDatastoreFileManager` so the existing file at that version is replaced in-place. The override is also forwarded through the Prefect flows in `performCalibration.py`.
- **Version bump**: `pyproject.toml` → `6.3.0`.

### Source
- `src/mag_toolkit/calibration/calibrators/CalibrationJob.py` — return type `tuple[Path, Path]` → `Sequence[Path]`
- `src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py` — `OUTPUT_SUBFOLDER_NAME`, stale-output prevention, `output_version` in MATLAB command, explicit `output_dir` param on `_write_user_config`
- `src/imap_mag/cli/calibrate.py` — `_validate_version_number_override`, multi-file save loop, `version_number_override` parameter
- `src/imap_mag/io/file/VersionedPathHandler.py` — `allow_overwrite: bool = False`
- `src/imap_mag/io/file/CalibrationLayerPathHandler.py` — `get_equivalent_data_handler()` copies `allow_overwrite`
- `src/imap_mag/io/DatastoreFileManager.py` — `allow_overwrite` branch in version selection
- `src/imap_mag/io/DBIndexedDatastoreFileManager.py` — `allow_overwrite` branch soft-deletes conflicting DB record before insert
- `src/prefect_server/performCalibration.py` — `version_number_override` forwarded in both flows

### Tests
- `tests/unit/test_calibrate_version_override.py` *(new)* — validation unit tests + end-to-end override tests
- `tests/unit/test_calibration_multi_file.py` *(new)* — extra files saved, zero files raises, unknown type raises
- `tests/unit/test_scripted_l2_calibration.py` — updated for `outputs/` subfolder, list return type, `output_version` in command
- `tests/unit/test_DatastoreFileManager.py` — `test_allow_overwrite_replaces_same_version` added
- `tests/unit/test_calibration_without_matlab.py` — `test_gradiometry_returns_list` added

---
_Generated by [Claude Code](https://claude.ai/code/session_01DyoiMch8Y1LpYXABhHWFKg)_

### DIFF
--- a/src/imap_mag/io/file/CalculatedOffsetsPathHandler.py
+++ b/src/imap_mag/io/file/CalculatedOffsetsPathHandler.py
@@ -3,7 +3,6 @@ import re
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import ClassVar
 
 from imap_mag.io.file.VersionedPathHandler import VersionedPathHandler
 
@@ -15,6 +14,14 @@ SPIN_PLANE = "spin_plane"  # spin-averaged offsets
 SPIN_OPTIMISED = "spin_optimised"  # spin-tone-corrected ("optimised") offsets
 OFFSET_TYPES = (SPIN_PLANE, SPIN_OPTIMISED)
 
+# Each offset type has its own file-name descriptor so the two products are uniquely
+# named (and not just distinguished by their folder). This mirrors what MATLAB writes.
+_DESCRIPTOR_BY_OFFSET_TYPE = {
+    SPIN_PLANE: "spin-plane-offsets",
+    SPIN_OPTIMISED: "spin-plane-optimised-offsets",
+}
+_OFFSET_TYPE_BY_DESCRIPTOR = {v: k for k, v in _DESCRIPTOR_BY_OFFSET_TYPE.items()}
+
 
 @dataclass
 class CalculatedOffsetsPathHandler(VersionedPathHandler):
@@ -22,11 +29,11 @@ class CalculatedOffsetsPathHandler(VersionedPathHandler):
 
     These live in ``calibration/calculated_offsets/<offset_type>/`` where
     ``offset_type`` is ``spin_plane`` (spin-averaged offsets) or ``spin_optimised``
-    (spin-tone-corrected offsets). The scripted-L2 MATLAB calibration writes one CSV
-    per sensor per day into each folder. The file name is identical in both folders,
-    so the folder (``offset_type``) is what distinguishes the two products.
+    (spin-tone-corrected offsets). Each offset type has a distinct file-name
+    descriptor so the two products are uniquely named:
 
-    File names look like ``imap_mag_mago-spin-plane-offsets_20260114_v024.csv``.
+    - ``spin_plane``     -> ``imap_mag_<sensor>-spin-plane-offsets_<date>_vNNN.csv``
+    - ``spin_optimised`` -> ``imap_mag_<sensor>-spin-plane-optimised-offsets_<date>_vNNN.csv``
     """
 
     mission: str = "imap"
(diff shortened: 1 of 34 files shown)


src/imap_mag/io/file/CalculatedOffsetsPathHandler.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+
+    mission: str = "imap"
+    instrument: str = "mag"
+    sensor: str | None = None  # "mago" or "magi"
+    offset_type: str | None = None  # one of OFFSET_TYPES
+    content_date: datetime | None = None  # date the offsets belong to
+    extension: str = "csv"
+
+    DESCRIPTOR: ClassVar[str] = "spin-plane-offsets"
```
alastairtree: @mhairifin this hardcodes the file descriptor insode the file name to never use "optimised" i think - i lifted this code from your PR. Am i right that this needs fixing?

mhairifin: Yes - see https://github.com/ImperialCollegeLondon/imap-pipeline-core/pull/350/changes#diff-bd13d6441503d3e2a5047ef1d22c9773f40093d93229a1dcb43aaa2c4d4f0f18 which has descriptor by offset type to include spin-plane-optimised-offsets

src/mag_toolkit/calibration/MatlabWrapper.py
```diff
@@ -185,6 +191,16 @@ def call_matlab(
                 line = line[len("CRITICAL: ") :]
                 log_method = logger.critical
 
+            # Capture the MATLAB output after the first "ans =" line, which is the start of the function's return value.
```
mhairifin: This is not always true. e.g. the below snippet would output `ans` twice. 
```
function ret = example(a)
    a+9
    ret = a+11;
end
```
```
>> example(5)

ans =

    14


ans =

    16
```
`ans` is just the generic term for uncaptured output in MATLAB. This is probably okay for now as this is just being used for logging but in general if we want to capture MATLAB output I think we might need a more robust solution - or always pass important information in files.

mhairifin (approved):
Small comment on logging but think it is ultimately inconsequential - overall looks good - pending changes to OffsetsPathHandler we discussed

src/mag_toolkit/calibration/MatlabWrapper.py
alastairtree: I wondered if you would spot this little easter egg! Yeah i was playing with the idea and as you say not using it currently, just logging. Good point though! Have tweaked it to only record the last "ans =" section now so might still be useful? Anyway, lets merge it regardless

src/imap_mag/io/file/CalculatedOffsetsPathHandler.py
alastairtree: i have integrated these changes! thanks. My version ditches the from_work_folder version - it just ensures the file is always in a correctly named folder or is ignored by the handler as the existing from_filename path always includes the relative path

PR #367, opened 2026-08-07
feat: add create_offsets flag to scripted-L2 calibration config

alastairtree:
## Summary

Adds a `create_offsets` flag to `ScriptedL2CalibrationConfig` and threads it through to the MATLAB `calibrate_l2_offsets()` call.

### Semantics

| Value | Effect |
|---|---|
| `None` (default, **automatic**) | `true` for normal mode, `false` for burst mode |
| `True` | always pass `create_offsets=true` regardless of mode |
| `False` | always pass `create_offsets=false` regardless of mode |

The automatic default reflects that burst mode rarely spans a long enough interval to produce meaningful offsets, so suppressing them there is the safe fallback.

## Changes

- `src/mag_toolkit/calibration/CalibrationConfig.py` — add `create_offsets: bool | None = None` field to `ScriptedL2CalibrationConfig`.
- `src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py`:
  - In `run_calibration()`: resolve the automatic default from the science mode before building the command.
  - In `_build_matlab_command()`: accept and emit `create_offsets=true|false` as a MATLAB named parameter; also update docstring for `produce_report` (was missing) and `create_offsets`.
- `tests/unit/test_calibration_config.py` — three tests for the new field (default None, explicit True, explicit False).
- `tests/unit/test_scripted_l2_calibration.py` — four tests in `TestCreateOffsets` covering the automatic-norm, automatic-burst, explicit-True-burst-override, and explicit-False-norm-override scenarios.

> **Stacked on:** #366 (fix: forward rotation and reference_frames in calibrate-and-apply split-by-day)

---
_Generated by [Claude Code](https://claude.ai/code/session_01CtFxoAUgR9SLW5PTchAGJX)_


src/mag_toolkit/calibration/CalibrationConfig.py
```diff
@@ -66,6 +66,7 @@ class ScriptedL2CalibrationConfig(CalibrationConfig):
     datastore_access_mode: DatastoreAccessMode = DatastoreAccessMode.READ_DIRECTLY
     matlab_repo: str
     produce_report: bool = False
+    create_offsets: bool | None = None
```
alastairtree: Make a custom enum for this so you choose explicitly between automatic, yes or no. No none option

alastairtree: Done — replaced `bool | None` with a `CreateOffsets` enum (`AUTOMATIC` / `YES` / `NO`) defined in `CalibrationDefinitions.py`. Also bumped the version to 6.4.0.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #380, opened 2026-08-14
Add Apache Parquet layer file support and calibrate-convert flow

alastairtree:
## Summary
- `calibrate`, `apply`, and `calibrate-and-apply` now accept a `layer_data_format` option (`csv` or `parquet`, defaulting to `parquet` with zstd compression) for the calibration layer's companion data file. Apply reads any mix of csv/parquet layers transparently.
- New `calibrate-convert` flow/CLI command converts existing layer files between csv, parquet, and cdf, verifying the converted data matches before publishing and (optionally) removing the superseded files.
- CSV epoch formatting was fixed to preserve nanosecond precision (previously silently truncated to microseconds), so csv and parquet round-trip identically.


src/imap_mag/cli/calibrate.py
```diff
@@ -166,10 +168,13 @@ def _save_calibration_outputs(
         if companion_path:
             # Add the companion data file to the output manager with its equivalent data handler, ensuring both files are saved together and co-versioned.
             # we deliberately did not save the data file with its own handler, as that would cause it to be versioned independently of the JSON layer.
-            outputManager.add_file(
-                companion_path,
-                path_handler=handler.get_equivalent_data_handler(),  # type: ignore
-            )
+            # The handler's own data_extension reflects the *requested* layer_data_format,
+            # which may not match what the calibration job actually produced (e.g. a
+            # scripted-L2 run that only writes csv); use the companion file's real
```
alastairtree: You can and should assume that if you tell scripted-L2 to produce parquet files it will not output layer data CSV file (in fact you can error if that happens) and it will produce the file type it was asked to produce

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,331 @@
...
+        typer.Option(help="Whether to save locally only or to also save to database"),
+    ] = SaveMode.LocalAndDatabase,
+) -> list[Path]:
+    """
+    Convert existing calibration layer files between csv, parquet and cdf formats.
+
+    Layers are found the same way as ``apply`` (exact filenames or glob patterns),
+    optionally restricted to a date range and/or science mode.
+
+    e.g. imap-mag convert --input-layers '*noop*' --output-layer-data-format csv
```
alastairtree:
```suggestion
    e.g. imap-mag convert --input-layers '*noop*.parquet' --output-layer-data-format csv
```

src/imap_mag/cli/convert.py
alastairtree, line 1: Add some helpful debug and info level logging to this command

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,331 @@
...
+        allow_overwrite=is_overwrite,
+    )
+
+    layer.version_major = output_handler.version_major
+    layer.version = output_handler.version
+
+    new_primary_path = work_folder / output_handler.get_filename()
+    new_companion_path: Path | None = None
+
+    if is_cdf_target:
```
alastairtree: Would be better if this type of layer in cdf file specific logic could be pushed into the calibration layer class

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,331 @@
...
+    versioned_source_path = datastore_finder.find_by_handler(
+        source_handler, throw_if_not_found=True
+    )
+    work_source_path = fetch_file_for_work(
+        versioned_source_path, work_folder, throw_if_not_found=True
+    )
+
+    old_datastore_paths: list[Path] = [versioned_source_path]
+    work_companion_path: Path | None = None
+    if source_handler.extension != FileType.CDF.value:
```
alastairtree: Would be better if this type of layer in cdf file specific logic could be pushed into the calibration layer class

src/imap_mag/io/FileFinder.py
```diff
@@ -204,6 +204,88 @@ def find_layers_by_date_and_patterns(
 
         return resolved
 
+    def find_layers_by_patterns(
```
alastairtree: Merge this logic with find_layers_by_date_and_patterns - do not duplicate code like this

src/prefect_server/performCalibration.py
alastairtree, line 1: This file is way too big now - split the flows into one file per flow

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -28,6 +31,7 @@ class CalibrationLayerPathHandler(VersionedPathHandler):
     extra_descriptor: str = ""
     content_date: datetime | None = None  # date data belongs to
     extension: str = "json"
+    data_extension: str = "parquet"  # extension for the companion data file
```
alastairtree: This handler needing to encode the extension of another sibling file as a property is a code smell - it is confusing and error prone. The handlers are supposed to represent the paths logic and data structure of files as they are created or read, not know the logic of what the file represents/holds.

This has been added because get_equivalent_data_handler() needs it when it is called, but that method has become a muddle. Refactor get_equivalent_data_handler() to be 2 different code paths that are called depending on the use case. 
- If the layer file pair already exists on disk then the data file name is encoded in the json of the file. Any callers using get_equivalent_data_handler() should instead use an instance of CalibrationLayer and read the file to ascertain the datafile path. 
- Any callers using get_equivalent_data_handler() to create a new datafile must already know if the datafile is going to be csv or parquet or whatever extension because the user has defined the layer format, So creators of new data files should call a renamed method create_new_datafile_handler(format: LayerDataFormat) and pass in the extension of the new file to be created using this new handler.

In summary - refactor this to make data_extension disappear.

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -69,7 +73,7 @@ def get_unsequenced_pattern(self) -> re.Pattern:
         )
 
     def is_metadata_file(self) -> bool:
-        """Determine if this handler represents a metadata JSON file (true) or a data csv/cdf/arrow type file (false)."""
+        """Determine if this handler represents a metadata JSON file (true) or a data csv/parquet/cdf type file (false)."""
         return self.extra_descriptor != "-data"
 
     def get_equivalent_data_handler(self) -> "CalibrationLayerPathHandler":
```
alastairtree: rename to create_new_datafile_handler and force callers to provide a LayerDataFormat as discussed above

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -130,6 +135,7 @@ def from_method(
         settings: "AppSettings",  # type: ignore  # noqa: F821
         mode: ScienceMode | None = None,
         version_number_override: tuple[int, int] | None = None,
+        layer_data_format: LayerDataFormat | None = None,
```
alastairtree: remove this param - and postpone the file type assignment until new method create_new_datafile_handler is called

src/imap_mag/io/file/CalibrationLayerPathHandler.py
alastairtree: prepare_for_version in CalibrationLayerPathhandler should push this logic into CalibrationLayer. CalibrationLayer should be responsible for changing the version of the layer and its data file as it understands the detail of the files best. Rename prepare_for_version to update_file_contents_based_on_version() and move this logic into CalibrationLayer. Try and simplify it based on the better info available inside that class.

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -266,12 +283,13 @@ def is_version_blocked_by_sibling(
         """
         if self.extension != "json":
             return False
+        new_companion = self._companion_csv_path(source_file)
```
alastairtree: self._companion_csv_path needs renaming to not be incorrect for a parquet layer data file

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -266,12 +283,13 @@ def is_version_blocked_by_sibling(
         """
         if self.extension != "json":
```
alastairtree, line 296: use is_metadata_file method instead of checking extension for json if possible

src/imap_mag/io/FileFinder.py
```diff
@@ -131,26 +131,54 @@ def find_layers_by_date_and_patterns(
...
+        """
+        return self.find_layers_by_patterns(
+            layers,
+            start_date=date,
+            end_date=date,
+            mode=mode,
+            throw_if_not_found=throw_if_not_found,
+        )
+
+    def find_layers_by_patterns(
```
alastairtree: Merge this method and find_layers_by_date_and_patterns - update all callers to find_layers_by_date_and_patterns provide a date range instead

src/imap_mag/io/FileFinder.py
```diff
@@ -161,49 +189,79 @@ def find_layers_by_date_and_patterns(
...
+
+        if start_date is None:
+            return [p for p in layers_folder.rglob("*") if p.is_file()]
+
+        months: dict[tuple[int, int], None] = {}
+        current = start_date.replace(day=1)
+        effective_end = end_date or start_date
+        while current <= effective_end:
+            months[(current.year, current.month)] = None
+            current = (current.replace(day=28) + timedelta(days=4)).replace(day=1)
```
alastairtree: Is this really the easiest way to iterate through months? Try and make __candidate_layer_paths much simpler.easier to read

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -148,6 +154,9 @@ def from_method(
             versioning_mode=VersionedPathHandler.VersionMode.USER_OVERRIDE
             if version_number_override
             else VersionedPathHandler.VersionMode.MAX_VERSION_PLUS_ONE,
+            data_extension=layer_data_format.value
+            if layer_data_format is not None
+            else "parquet",
```
alastairtree: Avoid magic strings for "parquet" here - use the enum or a constant

src/mag_toolkit/calibration/calibrators/ScriptedL2Calibration.py
```diff
@@ -315,6 +316,7 @@ def _build_matlab_command(
         matlab_mode: str,
         produce_report: bool,
         write_offsets: bool,
+        layer_data_format: str = "parquet",
```
alastairtree: Avoid magic strings for "parquet" here - use the enum or a constant

src/mag_toolkit/calibration/CalibrationLayer.py
```diff
@@ -253,7 +305,10 @@ def save_calibration_layer(self, filepath, createDirectory, save_contents=True):
                 )
             data_file_path = filepath.parent / self.metadata.data_filename
             if save_contents:
-                self._write_to_csv(data_file_path, createDirectory)
+                if data_file_path.suffix == ".parquet":
```
alastairtree: Avoid magic strings for "parquet" here - use the enum or a constant

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+from mag_toolkit.calibration.CalibrationDefinitions import CONSTANTS, LayerDataFormat
+
+app = typer.Typer()
+
+logger = logging.getLogger(__name__)
+
+_CONVERTIBLE_FORMATS = (FileType.CSV, FileType.PARQUET, FileType.CDF)
+
+
+def convert(
```
alastairtree, line 29: Add a short comment above showing how to call the convert function on the cli. Also add the same to the readme.

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+    )
+
+    old_datastore_paths: list[Path] = [versioned_source_path]
+    work_companion_path: Path | None = None
+    if not CalibrationLayer.is_self_contained_format(source_handler.extension):
+        source_peek = CalibrationLayer.from_file(work_source_path, load_contents=False)
+        if source_peek.metadata.data_filename:
+            companion_versioned_path = (
+                versioned_source_path.parent / source_peek.metadata.data_filename.name
+            )
```
alastairtree: Push this logic into a method on the calibration layer class. Look for other code like this as it probably exists elsewhere too and reuse/move into a single method.

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+            work_companion_path = fetch_file_for_work(
+                companion_versioned_path, work_folder, throw_if_not_found=True
+            )
+            old_datastore_paths.append(companion_versioned_path)
+
+    source_extension = (
```
alastairtree: Make this a layer file method get_data_file_type() and return the enum value which can then be used to get the extension

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+        source_handler, throw_if_not_found=True
+    )
+    work_source_path = fetch_file_for_work(
+        versioned_source_path, work_folder, throw_if_not_found=True
+    )
+
+    old_datastore_paths: list[Path] = [versioned_source_path]
+    work_companion_path: Path | None = None
+    if not CalibrationLayer.is_self_contained_format(source_handler.extension):
+        source_peek = CalibrationLayer.from_file(work_source_path, load_contents=False)
```
alastairtree: This peek version is pointless given the whole layer is loaded a few lines later in the method anyway. Just load it at the start either way.

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+    versioning_mode = (
+        VersionedPathHandler.VersionMode.USER_OVERRIDE
+        if is_overwrite
+        else VersionedPathHandler.VersionMode.MAX_VERSION_PLUS_ONE
+    )
+
+    output_is_self_contained = CalibrationLayer.is_self_contained_format(
+        output_layer_data_format.value
+    )
+    output_handler = CalibrationLayerPathHandler(
```
alastairtree: Move this construction logic go into a builder method inside the handler so that changes to the handlers do not create the need to change the convert function

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+    )
+    new_companion_path = (
+        work_folder / companion_filename if companion_filename else None
+    )
+
+    logger.debug(
+        f"Writing converted layer {layer_name} to {new_primary_path.name}"
+        + (f" (+ {new_companion_path.name})" if new_companion_path else "")
+    )
+    layer.writeToFile(new_primary_path)
```
alastairtree: This method should be refactored to be snake case

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+def _delete_old_datastore_file(
+    path: Path, app_settings: AppSettings, save_mode: SaveMode
+) -> None:
+    """Remove a superseded layer file from the datastore (and soft-delete its DB record)."""
+    relative_path = path.relative_to(app_settings.data_store).as_posix()
+    logger.info(f"Deleting superseded layer file {relative_path} after conversion.")
+    path.unlink()
+
+    if save_mode == SaveMode.LocalAndDatabase and Database.get_environment_url():
+        db = Database()
```
alastairtree: A database object is created at the beginning of the covert method, it should be reused here by passing it through and not recreated.

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,365 @@
...
+    involves_cdf: bool,
+) -> None:
+    """Read back the original and converted layer contents and confirm they match
+    before the converted files are published and the originals removed.
+
+    CDF stores offsets/timedelta as single-precision (CDF_FLOAT), so a tolerant
+    comparison is used whenever CDF is on either side of the conversion; csv/parquet
+    conversions must match exactly, since both are lossless for the values involved.
+    """
+    original = CalibrationLayer.from_file(original_path, load_contents=True)
```
alastairtree: The original layer has already been loaded into memory once above in the convert method. That should be reused here by passing the layer object instead of loading the entire datafile into memory again.

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -69,19 +72,30 @@ def get_unsequenced_pattern(self) -> re.Pattern:
         )
 
     def is_metadata_file(self) -> bool:
-        """Determine if this handler represents a metadata JSON file (true) or a data csv/cdf/arrow type file (false)."""
+        """Determine if this handler represents a metadata JSON file (true) or a data csv/parquet/cdf type file (false)."""
         return self.extra_descriptor != "-data"
 
-    def get_equivalent_data_handler(self) -> "CalibrationLayerPathHandler":
+    def create_new_datafile_handler(
```
alastairtree, line 87: This is another example of building a layer file handler from another handler - try and combine/share logic in this with other instances

alastairtree (commented):
Needs some more changes

alastairtree:
@mhairifin i have reviewed and tested this and seems to work well, can you take a look. Apologies it got quite big in the process.

src/imap_mag/cli/apply.py
```diff
@@ -472,12 +467,24 @@ def cleanup_workfolder_after_apply(
...
 
-    # add the .csv version of all layer files to the cleanup list as well
+    # add the companion data file (CSV or Parquet) for each JSON layer to the cleanup list
     for layer_file in workLayers:
-        if layer_file.suffix == ".json":
-            corresponding_csv = layer_file.with_suffix(".csv")
-            if corresponding_csv.exists():
-                files_to_cleanup.append(corresponding_csv)
+        handler = CalibrationLayerPathHandler.from_filename(layer_file)
+        if handler is not None and handler.is_metadata_file():
```
mhairifin, line 473: Some of this error handling seems a bit unnecessary, or like it should error more thoroughly. A calibration layer should always have an associated data filename, and if the layer fails to load from a file then I don't think we should just catch the exception and move on - that seems like an issue that deserves at least an error message.

src/imap_mag/cli/calibrate.py
```diff
@@ -166,10 +169,20 @@ def _save_calibration_outputs(
         if companion_path:
             # Add the companion data file to the output manager with its equivalent data handler, ensuring both files are saved together and co-versioned.
             # we deliberately did not save the data file with its own handler, as that would cause it to be versioned independently of the JSON layer.
-            outputManager.add_file(
-                companion_path,
-                path_handler=handler.get_equivalent_data_handler(),  # type: ignore
+            # The calibration job is expected to honour the requested layer_data_format
+            # exactly; a mismatch means the job did not produce the format it was asked
+            # for, which is a bug worth failing loudly on rather than silently adapting to.
```
mhairifin:
```suggestion
            # for
```
Sometimes the french lad is a bit too chatty in its comments

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,359 @@
...
+
+    numeric_cols = [
+        CONSTANTS.CSV_VARS.OFFSET_X,
+        CONSTANTS.CSV_VARS.OFFSET_Y,
+        CONSTANTS.CSV_VARS.OFFSET_Z,
+        CONSTANTS.CSV_VARS.TIMEDELTA,
+    ]
+    for col in numeric_cols:
+        if col not in original_df.columns or col not in converted_df.columns:
+            continue
```
mhairifin, line 330: I think this should be an error, not a `continue`.

src/imap_mag/cli/convert.py
```diff
@@ -0,0 +1,359 @@
...
+            matches = np.array_equal(original_values, converted_values, equal_nan=True)
+        if not matches:
+            raise ValueError(
+                f"Conversion verification failed for {layer_name}: "
+                f"column '{col}' values differ after conversion."
+            )
+
+    for col in [CONSTANTS.CSV_VARS.QUALITY_FLAG, CONSTANTS.CSV_VARS.QUALITY_BITMASK]:
+        if col not in original_df.columns or col not in converted_df.columns:
+            continue
```
mhairifin, line 352: This should also be an error I think - especially if it is in the original but not in the converted dataframe

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -69,21 +72,49 @@ def get_unsequenced_pattern(self) -> re.Pattern:
...
+        self, layer_data_format: LayerDataFormat
+    ) -> "CalibrationLayerPathHandler":
+        """Build a handler for a *new* companion data file in the given format.
+
+        Only for creating new output: the caller must already know what format
+        it is writing (csv/parquet), since that is a property of the write
+        operation, not something a path handler can infer. To find the
+        companion of an *existing* layer, read its data_filename from a
+        CalibrationLayer instance instead (see ``CalibrationLayer.from_file``).
```
mhairifin, line 95:
```suggestion
        Only for creating new output: the caller must already know what format
        it is writing (csv/parquet). To find the
        companion of an *existing* layer, read its data_filename from a
        CalibrationLayer instance instead (see ``CalibrationLayer.from_file``).
```

src/imap_mag/io/file/CalibrationLayerPathHandler.py
```diff
@@ -167,11 +199,18 @@ def _companion_csv_path(self, alongside: Path) -> Path:
...
-            if cal.metadata.data_filename:
-                return alongside.parent / cal.metadata.data_filename.name
+            companion = cal.get_datafile_path()
+            if companion is not None:
+                return companion
         except Exception:
             pass
-        return alongside.parent / self.get_equivalent_data_handler().get_filename()
+        # Last resort when the JSON can't be read to determine the real
+        # companion extension: guess the default. This path is only used for
```
mhairifin, line 207: I don't think this should guess the default - I think it should fail if it can't work it out from the file. Or at least debug log it is guessing? I think this is unintuitive behaviour, and you have a bigger problem if the JSON can't be read and understood

src/imap_mag/io/FileFinder.py
```diff
@@ -161,49 +169,80 @@ def find_layers_by_date_and_patterns(
...
+                if mode is not None and f"{mode.short_name}-layer" not in path.name:
+                    continue
+                if start_date is not None and handler.content_date is not None:
+                    effective_end = end_date or start_date
+                    if not (
+                        start_date.date()
+                        <= handler.content_date.date()
+                        <= effective_end.date()
+                    ):
+                        continue
```
mhairifin, line 195: This is partially a preference - but I think it can make code difficult to read when the control of the loop is managed heavily by many continue statements, and I think in this case these lines are checking if the candidate is a match and that might sit better in a small function

src/imap_mag/io/FileFinder.py
```diff
@@ -161,49 +169,80 @@ def find_layers_by_date_and_patterns(
...
+        # Represent each month as a single "months since year 0" integer so the
+        # range can be walked with plain integer arithmetic (no calendar edge
+        # cases like day-31/leap-year rollover to worry about).
+        effective_end = end_date or start_date
+        start_month_index = start_date.year * 12 + start_date.month
+        end_month_index = effective_end.year * 12 + effective_end.month
+
+        candidate_paths: list[Path] = []
+        for month_index in range(start_month_index, end_month_index + 1):
+            year, month = divmod(month_index - 1, 12)
```
mhairifin, line 240: It does feel like there should be a way to just "walk the months" over the year boundaries so to speak but I can see that it's a weirdly complex problem, and this is a good solution

src/mag_toolkit/calibration/CalibrationLayer.py
```diff
@@ -269,22 +355,148 @@ def save_calibration_layer(self, filepath, createDirectory, save_contents=True):
...
+            Path(self.metadata.data_filename).name
+            if self.metadata.data_filename
+            else None
+        )
+        if current is None:
+            return source_file
+
+        # Preserve the companion's actual current format (csv/parquet) rather
+        # than guessing — we are re-versioning an existing file, not creating
+        # a new one, so its format is already fixed.
```
mhairifin:
```suggestion
        # Preserve the companion's current format (csv/parquet) 
```

src/prefect_server/calibrateAndApplyFlow.py
```diff
@@ -0,0 +1,200 @@
+from datetime import date, datetime
```
mhairifin, line 1: I am assuming all of these flows (other than `calibrate-and-convert`) are just refactors of existing flows rather than with new content other than passing the output layer format. Let me know if that is incorrect and I can more thoroughly review them

mhairifin (commented):
Overall looks good! I'd be interested to know if you have any numbers on how much smaller these files are, and comparatively how fast/slow parquet files of the same content read/write

src/prefect_server/calibrateAndApplyFlow.py
alastairtree: That's correct just trying to make it one flow per file!

src/imap_mag/io/FileFinder.py
alastairtree: Yeah good point, maybe an inline function for the continue logic and just one case calling continue?

